PR #114, opened 2025-07-18
feature: Idanl/feature/config file and docker file

idanlodzki:
## Issue Reference

---

## What Was Changed

---

## Why Was It Changed

---

## Screenshots (Optional)

---


scripts/docker-setup.sh
```diff
@@ -0,0 +1,196 @@
...
+        print_status "Found Docker Compose (v2)"
+    fi
+else
+    DOCKER_COMPOSE_CMD="docker-compose"
+    print_status "Found Docker Compose (v1)"
+fi
+
+# Create data directories
+print_info "Creating data directories..."
+mkdir -p data/database
```
yuvalmich: we dont use the database in here

yuvalmich: We use the database currently in `apps/server/service_peek.db`

I saw that you added a mount on the db directory which is great, but either move the db path also to the new location or change the mount because the code and the mount are directed to a different location

apps/server/src/config/config.ts
```diff
@@ -0,0 +1,115 @@
...
+    server: {
+      port: 3001,
+      host: 'localhost'
+    },
+    client: {
+      port: 8080,
+      api_url: 'http://localhost:3001/api/v1'
+    },
+    database: {
+      path: '../../service_peek.db'
```
yuvalmich, line 92: also this

PR #160, opened 2025-08-11
fix default config

idanlodzki:
Issue Reference
Fixes development setup friction - eliminates manual directory creation for new developers

What Was Changed
Added automatic directory creation during server startup:

Database directory: Auto-creates 
/apps/server/data/database/
 in 
initializeDb()
Private keys directory: Auto-creates 
/apps/server/data/private-keys/
 via new 
initializePrivateKeysDir()
Configuration: Updated paths to co-locate both directories under 
/apps/server/data/
Files modified:

/apps/server/src/dal/db.ts
 - Added directory creation before SQLite connection
/apps/server/src/dal/sshClient.ts
 - Added private keys directory initialization
/apps/server/src/index.ts
 - Integrated directory setup into server startup
/apps/server/src/config/config.ts
 - Updated default paths for consistency
Why Was It Changed
Before: ❌ npm run dev failed with "directory does not exist" errors requiring manual setup

After: ✅ npm run dev works immediately - directories created automatically

Benefits:

Zero-configuration developer onboarding
Eliminates startup errors
Consistent data organization
Docker compatibility maintained
Additional Context (Optional)
Both directories now use the same path pattern (../../data/) and are created in 
/apps/server/data/
 for organized data management. Includes proper logging when directories are created.

Feedback submitted


apps/server/src/config/config.ts
```diff
@@ -61,10 +61,10 @@ function getDefaultConfig(): OpsimateConfig {
             api_url: 'http://localhost:3001/api/v1'
         },
         database: {
-            path: '../../data/opsimate.db'
+            path: '../../data/database/opsimate.db'
```
yuvalmich, line 64: is it on purpose? it changes the existing db location

apps/server/src/dal/sshClient.ts
```diff
@@ -9,9 +9,23 @@ const logger = new Logger('dal/sshClient');
 
 function getPrivateKeysDir(): string {
     const securityConfig = getSecurityConfig();
-    return path.isAbsolute(securityConfig.private_keys_path)
+    const privateKeysPath = path.isAbsolute(securityConfig.private_keys_path)
         ? securityConfig.private_keys_path
-        : path.resolve(__dirname, '../../../../', securityConfig.private_keys_path);
+        : path.resolve(__dirname, securityConfig.private_keys_path);
```
yuvalmich, line 14: I wonder if there is a way to make it relative to the project root and not to the current dir, because if we refactor and move this file to another folder it might stop working

PR #259, opened 2025-09-24
Review

yjrathod:


apps/server/src/dal/secretsMetadataRepository.ts
```diff
@@ -78,5 +78,47 @@ export class SecretsMetadataRepository {
             `).run();
         });
     }
+    async updateSecretOnserver(id: number, data: Partial<{
```
Sahar541998: No need the "server" suffix we are on the server.

```suggestion
    async updateSecret(id: number, data: Partial<{
```

apps/client/src/lib/api.ts
```diff
@@ -588,6 +589,68 @@ export const secretsApi = {
       };
     }
   },
+
+    updateSecretOnServer : async (secretId: number, updateData: {
```
Sahar541998: not need for the suffix `OnServer`
```suggestion
    updateSecret : async (secretId: number, updateData: {
```

apps/client/src/lib/api.ts
```diff
@@ -588,6 +589,68 @@ export const secretsApi = {
...
+      const token = localStorage.getItem('jwt');
+      
+      const response = await fetch(url, {
+        method: 'PATCH',
+        headers: {
+          ...(token ? { 'Authorization': `Bearer ${token}` } : {}),
+        },
+        credentials: 'include',
+        body: formData,
+      });
```
Sahar541998: dont use `fetch` use `apiRequest` 
see how deleteSecret implemented

apps/client/src/lib/api.ts
```diff
@@ -588,6 +589,68 @@ export const secretsApi = {
...
+        try {
+          const errorJson = JSON.parse(errorText);
+          return {
+            success: false,
+            ...errorJson,
+          };
+        } catch {
+          return {
+            success: false,
+            error: `HTTP ${response.status}: ${errorText || 'Unknown error'}`,
```
Sahar541998: should be like this - 
```suggestion
            error: error instanceof Error ? error.message : 'Unknown error occurred'
```

to be aligned

apps/client/src/lib/sslKeys.ts
```diff
@@ -42,3 +42,60 @@ export async function deleteSecretOnServer(secretId: number): Promise<{ success:
...
+        });
+
+        if (!response.ok) {
+            // Handle 401 unauthorized (same as your main API)
+            if (response.status === 401) {
+                window.location.href = "/login?expired=true";
+                localStorage.removeItem('jwt');
+                return;
+            }
+
```
Sahar541998: this is duplicated code. you need to just use the `secretsApi`
for example `deleteSecretOnServer` just call for `secretsApi.deleteSecret(secretId);` and wrap it with try catch

apps/server/src/dal/secretsMetadataRepository.ts
```diff
@@ -78,5 +78,47 @@ export class SecretsMetadataRepository {
...
+            return false; // Nothing to update
+        }
+
+        // Add the id parameter for WHERE clause
+        values.push(id);
+
+        // FIXED: Table name was 'secrets_metadata', should be 'secrets'
+        const sql = `UPDATE secrets SET ${fields.join(', ')} WHERE id = ?`;
+        
+        const result = this.db.prepare(sql).run(...values); // FIXED: spread the values array
```
Sahar541998: use `runAsync` and not `run`

apps/server/src/dal/secretsMetadataRepository.ts
```diff
@@ -78,5 +78,47 @@ export class SecretsMetadataRepository {
...
+        
+        return result.changes > 0;
+    } catch (error) {
```
Sahar541998: don't catch here. since we are returning promise

apps/server/src/bl/secrets/secretsMetadata.bl.ts
```diff
@@ -43,6 +43,59 @@ export class SecretsMetadataBL {
...
+            // If fileName is being updated, verify the new file exists
+            if (data.fileName && data.fileName !== secret.fileName) {
+                const fs = await import('fs');
+                const oldPath = path.resolve(getSecurityConfig().private_keys_path, secret.fileName);
+                const newPath = path.resolve(getSecurityConfig().private_keys_path, data.fileName);
+
+                // Check if new file exists
+                if (!fs.existsSync(newPath)) {
+                    throw new Error(`New secret file not found: ${data.fileName}`);
+                }
```
Sahar541998: move this logic to private func

apps/server/src/bl/secrets/secretsMetadata.bl.ts
```diff
@@ -43,6 +43,59 @@ export class SecretsMetadataBL {
...
+                    try {
+                        fs.unlinkSync(oldPath);
+                        logger.info(`Successfully deleted old secret file: ${oldPath}`);
+                    } catch (fileError) {
+                        const error = fileError as Error;
+                        logger.error(`Error deleting old secret file: ${oldPath}`, error);
+                        throw new Error(`Failed to remove old secret file: ${error.message}`);
+                    }
+                }
+            }
```
Sahar541998: move this logic to private func

apps/server/src/api/v1/secrets/controller.ts
```diff
@@ -69,4 +78,77 @@ export class SecretsController {
...
+        updateSecretOnServer = async (req: Request, res: Response) => {
+    try {
+        const secretId = parseInt(req.params.id);
+        if (isNaN(secretId)) {
+            return res.status(400).json({success: false, error: 'Invalid secret ID'});
+        }
+
+        // Validate request body - only validate if there are body params
+        let updateData: any = {};
+        if (Object.keys(req.body).length > 0) {
```
Sahar541998: move this logic to a variable 
const isEmptyBody = Object.keys(req.body).length === 0
but are you sure you cant just call `UpdateSecretSchema.parse(req.body);` 
what happen if there is no body?

apps/server/src/api/v1/secrets/controller.ts
```diff
@@ -69,4 +78,77 @@ export class SecretsController {
...
+
+        // Prepare data for update - FIXED TYPE
+        const updatePayload: Partial<{
+            name: string;
+            type: SecretType;
+            fileName: string;
+        }> = {};
+
+        if (updateData.displayName) updatePayload.name = updateData.displayName;
+        if (updateData.secretType) updatePayload.type = updateData.secretType;
```
Sahar541998, line 103: dont create and then assign its better to do it in 1 line
```
const updatePayload = {
 displayName:updateData.displayName,
type: updateData.secretType
} 
```

Sahar541998 (changes requested):
wrote small comments!
nice job

tamirg (commented):
Can you please add a description of what this PR is about to solve/implement?
it will be helpful for reviewers to add the issue link, short meaningful description at the title, screenshot of main changes/additions in ui in the description etc

Issue #263, opened 2025-09-26
[Bug] Refresh provider doesn't work


yuvalmich:
**Describe the bug**

Refresh provider doesn't work

<img width="476" height="470" alt="Image" src="https://github.com/user-attachments/assets/77b213c8-63f2-421d-8bf0-7fb9210667fc" />

**To Reproduce**
1. Click refresh as described in the image above

**Expected behavior**
The provider data needs to be updated from the remote, including the statuses

But we get 404 instead

VedantAnand17:
is this taken?

idanlodzki:
It’s currently available - you can take it!
Please make sure to join our Slack community and star the repository.
Also, keep your code clear and include screenshots in your PR.

stockman2836:
Can i start it please?

idanlodzki:
Sure

PR #271, opened 2025-10-01
[FIX] Reload provider not functioning

tamirg:
## Issue Reference

[Issue #263](https://github.com/OpsiMate/OpsiMate/issues/263)
---

# What Was Changed

### Backend
Added Provider Refresh Endpoint - New POST /api/v1/providers/:providerId/refresh endpoint that performs real-time SSH/K8s connection to discover services and update their status
Enhanced Provider Business Logic - Added refreshProvider() method that connects to remote servers, discovers services with live status, and updates the database
Router Update - Added refresh route while maintaining existing GET endpoint for cached data

### Frontend
Updated Refresh Functionality - Modified handleRefreshProvider() to use the new real-time refresh endpoint with loading states and improved user feedback
Client API - Added refreshProvider() method for real-time data fetching
Empty Status Handling - Added fallback logic to display "unknown" (gray badge) when serviceStatus is empty or missing across all components

---

### Problem: The "Refresh" button only fetched cached database data without connecting to servers, showing stale information. Additionally, services with no pods displayed blank status badges.
Solution:
Real-time refresh endpoint that actually SSH/connects to servers for accurate current status
Distinguishes between fast cached reads (GET) and real-time discovery (POST /refresh)
Consistent "unknown" status display prevents blank badges and improves UX
Impact: Users now get accurate real-time service status on demand with better feedback and consistent status display across the UI.

---

## Screenshots (Optional)

---

### DIFF
--- a/apps/server/src/api/v1/providers/controller.ts
+++ b/apps/server/src/api/v1/providers/controller.ts
@@ -33,30 +33,6 @@ export class ProviderController {
         }
     }
 
-    async getProviderById(req: Request, res: Response) {
-        try {
-            const providerId = parseInt(req.params.providerId);
-            if (isNaN(providerId)) {
-                return res.status(400).json({success: false, error: 'Invalid provider ID'});
-            }
-
-            const provider = await this.providerBL.getProviderById(providerId);
-            if (!provider) {
-                return res.status(404).json({success: false, error: 'Provider not found'});
-            }
-
-            delete provider.password;
-            res.json({success: true, data: provider});
-        } catch (error) {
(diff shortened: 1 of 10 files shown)


apps/server/src/api/v1/providers/controller.ts
```diff
@@ -33,6 +33,60 @@ export class ProviderController {
         }
     }
 
+    async getProviderById(req: Request, res: Response) {
```
yuvalmich: I probably missed it but is it used by the client side?

PR #273, opened 2025-10-01
[FIX]: Update provider IP validation to accept DNS hostnames as well as IP addresses

### LINKED ISSUE #200: [Bug] VM Provider Validation Fails on Hostnames – Only Accepts IPs
Describe the bug
When configuring a provider using a DNS hostname instead of an IP address, the validation incorrectly throws an error saying the IP is not valid.

To Reproduce

Go to add/edit a provider

Enter a valid hostname (e.g., my-server.example.com)

See a validation error expecting an IP

Expected behavior
The validator should accept both IP addresses and valid hostnames.

Additional context
Prevents using providers behind DNS, which is common in dynamic or cloud-based environments.

BhavyaMPatel:
## Issue Reference

Closes #200 

---

## What Was Changed

provider IP validation logic was updated to accept both IPv4 addresses and DNS hostnames
using hostname regex.

---

## Why Was It Changed

so, providers can be configured using DNS hostnames

---

## Screenshots (Optional)

---

hostname regex can we validate here : https://regex101.com/r/6Dnwyr/1

### DIFF
--- a/apps/client/src/components/ProviderSidebar.tsx
+++ b/apps/client/src/components/ProviderSidebar.tsx
@@ -20,13 +20,15 @@ import { SecretMetadata } from "@OpsiMate/shared";
 
 // --- FORM SCHEMAS ---
 
+const ipRegex = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/;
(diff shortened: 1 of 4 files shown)


BhavyaMPatel:
requesting to validate hostname regex as well : https://regex101.com/r/6Dnwyr/1

idanlodzki:
same issue please check it - go to "myprovider page and try to **edit** the provider info
@BhavyaMPatel

BhavyaMPatel:
Screenshot (55)

Screenshot (56)

edir provider info works fine

Issue #251, opened 2025-09-22
[Bug] provider validation check - handle duplications


NizCom:
**Describe the bug**
A user can create more than one provider of the same type with the same name.

**To Reproduce**

Go to the "Add Provider" page.

Click the "Configure" button (either Server or Kubernetes Cluster) and create a provider using a name such as "Test."

Repeat the previous action.

You will observe that there are two providers of the same type with identical names.

**Expected behavior**
When attempting to create another provider of the same type with an existing name, the application should display an error and prevent the creation of the duplicate provider.

**Additional context**
None.

PR #280, opened 2025-10-02
Fix: providers names check before creation - no duplications allowed

NizCom:
## Issue Reference
https://github.com/OpsiMate/OpsiMate/issues/251

## What Was Changed
Duplicate provider names are now prevented during creation. If a user tries to create a provider with a name that already exists for the same type, the application displays an error and does not allow the duplicate provider to be created.

## Why Was It Changed

Allowing multiple providers of the same type with identical names caused confusion and issues when managing providers. This change ensures each provider name is unique within its type, improving usability and preventing potential misconfigurations.

## Screenshots (Optional)

Screenshot 2025-10-02 232735

None.

### DIFF
--- a/apps/server/src/api/v1/providers/controller.ts
+++ b/apps/server/src/api/v1/providers/controller.ts
@@ -46,7 +46,8 @@ export class ProviderController {
                 res.status(400).json({success: false, error: 'Validation error', details: error.errors});
             } else {
                 logger.error('Error creating provider:', error);
-                res.status(500).json({success: false, error: 'Internal server error'});
+                const errorMessage = error instanceof Error ? error.message : 'Internal server error';
+                res.status(400).json({success: false, error: errorMessage});
             }
         }
     }
--- a/apps/server/src/bl/providers/provider.bl.ts
+++ b/apps/server/src/bl/providers/provider.bl.ts
@@ -33,6 +33,12 @@ export class ProviderBL {
         try {
             logger.info(`Starting to create provider`, { extraArgs: { ...providerToCreate } });
             
+            // Check if a provider with the same name already exists
+            const existingProvider = await this.providerRepo.getProviderByName(providerToCreate.name);
+            if (existingProvider) {
+                throw new Error(`A provider with the name '${providerToCreate.name}' already exists`);
+            }
+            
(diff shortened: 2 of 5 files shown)


packages/shared/src/schemas.ts
```diff
@@ -3,7 +3,6 @@ import {IntegrationType, ProviderType, ServiceType, Role, SecretType} from './ty
 
 export const CreateProviderSchema = z.object({
     name: z.string().min(1, 'Provider name is required'),
-    providerIP: z.string().ip('Invalid IP address').optional(),
```
yuvalmich, line 6: Why this and the other types was changed?

apps/server/src/bl/providers/provider.bl.ts
```diff
@@ -33,6 +33,12 @@ export class ProviderBL {
         try {
             logger.info(`Starting to create provider`, { extraArgs: { ...providerToCreate } });
             
+            // Check if a provider with the same name already exists
+            const existingProvider = await this.providerRepo.getProviderByName(providerToCreate.name);
```
yuvalmich, line 37: I would implement it differently to reduce extra db call and make it atomic.

Once you've implemented unique check via the db, you can try and save it and if it fails, handle it accordingly.

This way you avoid race condition issues and reduce latency

yuvalmich: Apperantly this issue happens a lot in this project so you can skip it for now and I'll fix it widely

apps/server/src/api/v1/providers/controller.ts
```diff
@@ -46,7 +46,8 @@ export class ProviderController {
                 res.status(400).json({success: false, error: 'Validation error', details: error.errors});
             } else {
                 logger.error('Error creating provider:', error);
-                res.status(500).json({success: false, error: 'Internal server error'});
+                const errorMessage = error instanceof Error ? error.message : 'Internal server error';
+                res.status(400).json({success: false, error: errorMessage});
```
yuvalmich, line 50: Why was it changed to 400?

We will return now always 400 also in general errors

We need to distinguish your specific new error

PR #313, opened 2025-10-04
fixed #279

### LINKED ISSUE #279: [Bug] Service Actions should be disabled in some cases (running status => disabled Start button)
**Describe the bug**
When service is already in "running" state, "start" actions should not be available:

Image

Image

(Same for stopped service, "stop" action should not be allowed)

**To Reproduce**
Run a service in your environment
click on the service in both pages My Providers and Dashboard and see the bug.

**Expected behavior**
The buttons should be disabled (if the service in running state - start should be disabled, stopped - stop should be disabled)

Lucky-Malik:
Issue Reference
Improves service action button UX by preventing invalid operations

What Was Changed

Added disabled state logic to Start, Stop, and Restart buttons in both Dashboard and MyProviders pages
Service action buttons now intelligently disable based on current service status
All action buttons remain visible but are grayed out when not applicable

Why Was It Changed
Previously, users could attempt to start an already running service or stop an already stopped service, leading to unnecessary API calls and confusing UX. The new implementation:

Prevents invalid service operations
Provides clearer visual feedback about available actions
Maintains consistency across Dashboard and MyProviders pages
Improves overall user experience by showing what actions are possible

Additional Context (Optional)
Implementation details:

Added canStart, canStop, and canRestart helper functions/memos
Buttons check service status (running/stopped/offline/error) before enabling
Works with both single and multiple service selections
Logic handles edge cases like services in error states

Technical notes:

Dashboard uses useMemo hooks for performance optimization with multi-select
MyProviders uses simpler helper functions since it handles single services
Both implementations follow the same business logic for consistency

### DIFF
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -1,41 +1,62 @@
+// src/App.tsx
+import React from "react";
 import { Toaster } from "@/components/ui/toaster";
 import { Toaster as Sonner } from "@/components/ui/sonner";
 import { TooltipProvider } from "@/components/ui/tooltip";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { BrowserRouter, Routes, Route } from "react-router-dom";
-import { Providers, MyProviders, Integrations, NotFound, Register, Login, Settings, Dashboard, Profile, Alerts, TVMode } from "./pages";
+import {
+  Providers,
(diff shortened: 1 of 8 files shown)


idanlodzki:
hi @Lucky-Malik please add screenshots and add PR title

Issue #304, opened 2025-10-03
[Feature] To add a dark and light mode button on the left sidebar for easier access


Aman-agraw-35:
Currently the button for Dark/Light mode is in the profile. I want to make a seperate avatar for it on the left sidebar. It will make it easier for the user to change the mode from the main page itself without going to the profile menu. We will be keeping  the theme option in the profile too as it is now, but a seperate button will be there outside

<img width="1013" height="902" alt="Image" src="https://github.com/user-attachments/assets/f2937a63-38f6-41f3-84b4-531b9bb01f57" />

Assign it to me

Aman-agraw-35:
@idanlodzki  Someone assigned it to me. If you want, I can transfer that feature to the navabar or somewhere else on the webpage. I can understand that the button causes some change in the UX, but it will be nice to keep the dark/light mode feature to a place where it is clearly visible to the user. When I saw the theme inside the profile section, I realised the need of it to be displayed somewhere outside. Rest is on you. Thanks

idanlodzki:
Yes, you can try to change it to other place, let’s see if it’s works @Aman-agraw-35

Aman-agraw-35:
@idanlodzki Can I make the button above scroll-to-top button 

![Image](https://github.com/user-attachments/assets/79c50bd2-f487-4ba7-b832-fd5e9dcc64b6)

![Image](https://github.com/user-attachments/assets/de761a9e-437f-4612-aef7-cb563a7fb2ad)

I mean where the cursor points in the last image

idanlodzki:
Thanks for the effort in creating the button!
For now, I prefer to keep it as part of the user settings. Changing the mode doesn’t feel critical enough to have such a prominent place, and when I tried it, the user experience felt a bit less smooth.
Sorry about that, but I’ll go ahead and close the issue and the PR for now.

PR #318, opened 2025-10-04
[FEAT]: To add a dark and light mode button on the left sidebar for easier access, Solves  #304

Aman-agraw-35:
[Feature] To add a dark and light mode button on the left sidebar for easier access #304

### **Solve Issue #304 for hactober-fest-2k25 **

I have created a left theme toggle on the left sidebar for the convinience of user.

image
image
image
image

### DIFF
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -1,41 +1,71 @@
+// src/App.tsx
+import React from "react";
 import { Toaster } from "@/components/ui/toaster";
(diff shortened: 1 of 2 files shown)


idanlodzki:
Hey, thanks for bringing this up — I really appreciate it!
Unfortunately, it’s not needed at the moment, or at least not in this approach.

It hurts the user experience since each button opens a new page. I’d prefer not to have a sidebar button that only changes the mode. Note that we previously had one like that, but we removed it to keep a consistent and clean user experience.

PR #325, opened 2025-10-05
[FIX]: 

### LINKED ISSUE #289: [Bug] Cannot edit provider without entering password (SSH key flow broken)
Describe the bug
When editing a provider, the form requires a password even if the provider is configured to use an SSH key. This prevents updating provider details unless a password is supplied, which is inconsistent with the creation stage where SSH key is supported without a password.

To Reproduce

Go to Providers.

Edit an existing provider configured with SSH key authentication.

Try to save changes without entering a password.

Observe that the edit fails or requires a password.

Expected behavior
Editing a provider should work the same as creation:

If a provider uses an SSH key, password should not be required.

Only validate password if that method of authentication is selected.

Additional context

Works fine during provider creation (SSH key without password).

Editing logic seems to always validate for password regardless of auth type.

Likely a frontend form validation or backend API check issue.

yjrathod:
## Issue Reference
Closes #289

---

## What Was Changed
- Made password field conditionally required (only when no SSH key selected)
- Modified form submission to send only password OR secretId, not both
- Clear password when SSH key is selected

---

## Why Was It Changed
Edit provider form was always requiring password even for SSH key authentication, preventing users from updating providers configured with SSH keys.

---

## Additional Context
Password field now respects the existing placeholder text: "leave empty to use key file"

### DIFF
--- a/apps/client/src/components/EditProviderDialog.tsx
+++ b/apps/client/src/components/EditProviderDialog.tsx
@@ -83,7 +83,7 @@ export function EditProviderDialog({
         providerIP: provider.providerIP || "",
         username: provider.username || "",
         secretId: secretId,
-        password: provider.password || "",
+        password: "",
         SSHPort: provider.SSHPort || 22,
         providerType: provider.providerType || 'VM',
       });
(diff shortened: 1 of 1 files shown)


idanlodzki:
Thanks for the PR!
I think it would be better to implement it the same way we handle adding a new service.
In the edit view, I’d like to see an option to choose either a password or an SSH key - not both at the same time.

PR #338, opened 2025-10-06
[FEAT]: Add filteredLogs and date range filtering to audit log table

### LINKED ISSUE #296: [Feature] Make Audit Log searchable
Is your feature request related to a problem? Please describe.
The Audit Log currently lists events but does not provide a way to search or filter them. This makes it difficult to quickly find specific actions, especially as the number of logged events grows.

Describe the solution you'd like
Add search and filtering capabilities to the Audit Log UI, allowing users to:

Search by user, resource type (provider, service, integration, etc.), or action (create, update, delete).

Filter by time range.

Describe alternatives you’ve considered

Exporting logs manually and searching externally (inefficient and not user-friendly).

Scrolling through the UI (slow and error-prone).

Additional context

Improves usability of the Audit Log for admins.

Makes compliance checks, troubleshooting, and investigations much faster.

Future enhancements could include advanced filters and export options.

vansh-nagar:
## Issue Reference

Closes #296 

---

## What Was Changed

Added search and date range filtering to the Audit Log UI. Users can now:

- Search logs by **user**, **resource type**, or **action type**.  
- Filter logs by a **custom date range**.  
- See filtered results immediately in the UI without affecting the initial logs display on page load.

---

## Why Was It Changed

The Audit Log previously displayed all events without any filtering or search capabilities, making it hard to locate specific actions. This change improves usability, efficiency, and allows admins to quickly find relevant log entries for troubleshooting, compliance, or audits.

---

## Video

https://github.com/user-attachments/assets/dd796feb-bbb1-4565-bec9-ccd163f3601f

---

### DIFF
--- a/apps/client/src/pages/Settings.tsx
+++ b/apps/client/src/pages/Settings.tsx
@@ -96,9 +96,6 @@ const Settings: React.FC = () => {
 useEffect(() => {
   let result = logs;
 
-  console.log(dateRange);
-
-  // 1️⃣ Apply string filter first
   const query = searchAuditLogQuery?.toLowerCase().trim();
   if (query && selectedFilter) {
     result = result.filter(log => {


yuvalmich (commented):
The video looks really good!

apps/client/src/pages/Settings.tsx
```diff
@@ -60,13 +71,68 @@ const Settings: React.FC = () => {
...
+        setIsFiltering(filteredLogs.length > 0);
+    }, [filteredLogs]);
+
+
+useEffect(() => {
+  let result = logs;
+
+  console.log(dateRange);
+
+  // 1️⃣ Apply string filter first
```
yuvalmich: Please remove the emoji from the code

vansh-nagar:
> I think we may have a problem, are these results coming from backend as in api call? or sole frontend i think its mostly frontend based @yuvalmich take a look please this will cause errors on future scopes, i feel it should be a backend call giving response.

I checked all the searches on the settings page — every filter is being applied on the frontend.

KaranNegi20Feb:
Would you be willing to rework on it? like fix the backend call? with the apis stuff ? rather than frontend else we may have to revert back

yuvalmich:
@KaranNegi20Feb 
You are totally right, I missed the pagination issue.

Unfortunately we will need to revert it as it causes an important bug

Issue #216, opened 2025-09-10
[Bug] Alerts Remain After Tag or Tagged Service Is Removed


idanlodzki:
Describe the bug
When a tag is removed, or when a service with a tag is deleted, the related alerts continue to appear on the Alerts page instead of being cleared or updated.

To Reproduce
	1.	Add a tag to a service.
	2.	Trigger or view alerts for that tagged service.
	3.	Remove the tag or delete the tagged service.
	4.	Alerts remain visible on the Alerts page.

Expected behavior
Alerts tied to removed tags or deleted services should be cleared, updated, or marked as invalid.

Additional context
This causes confusion since users may think the service is still problematic even after it’s removed.

PR #342, opened 2025-10-07
[FIX]: Fix of remaining alerts

Oleksii101:
## Issue Reference

https://github.com/OpsiMate/OpsiMate/issues/216

---
## What Was Changed

Backend:
-Grafana pull job (pull-grafana-alerts-job.ts)

Fetch Grafana alerts and map each alert to every service that has the matching tag.

Create/update per-service alerts with composite IDs: id = fingerprint:serviceId.

Persist service_id with each alert.

-Database

Added a field : service_id INTEGER REFERENCES services(id) ON DELETE CASCADE to the table ALERTS .

insertOrUpdateAlert now accepts service_id.

toSharedAlert now returns serviceId (number | undefined).

New repo methods in alerts:

deleteAlertsByTag(tag)

deleteAlertsByService(serviceId)

deleteAlertsByServiceAndTag(serviceId, tag)

Tag repo additions:

countServicesUsingTag(tagId) for cleanup logic.

findServiceIdsByTagName(tagName) to map alerts to services.

-TagController:

When removing a tag from a service, delete alerts only for that (service, tag) pair.

If a tag is no longer used by any service, delete all alerts with that tag.

Frontend:

-Alerts page

New Service column. Shows service name for each alert.

getServiceId reads serviceId from the alert field or parses it from id (fingerprint:serviceId).

-Dashboard / TV Mode

Fixed alerts counting: only non-dismissed alerts and no duplicates per service.

---

## Why Was It Changed

Previously alerts were matched only by tag, which caused  no reliable way to associate an alert with a specific service.
--------
The new approach (id = fingerprint:serviceId + storing service_id) provides:

-When a tag is removed from a service, all alerts for that pair (service, tag) are deleted

-Automatic cascade delete when a service is removed.

-Correct service names on the Alerts page.

-When a tag is deleted entirely, all alerts associated with that tag are deleted

---

https://github.com/user-attachments/assets/3c656c80-aa25-40f8-879a-7ebfbf39aec0

### DIFF
--- a/apps/client/src/components/AddServiceDialog.tsx
+++ b/apps/client/src/components/AddServiceDialog.tsx
@@ -166,17 +166,30 @@ export function AddServiceDialog({ serverId, serverName, providerType, open, onC
     }
 
     setLoading(true);
-    const serviceData = {
-      providerId: parseInt(serverId),
-      name: serviceName,
-      serviceType: "SYSTEMD" as const,
-      serviceStatus: "unknown" as const
-    };
-
-    console.log('Creating systemd service with data:', serviceData);
-
     try {
+      const serverStatus = await providerApi.getProviderInstances(parseInt(serverId));
+      if (!serverStatus.success || !serverStatus.data) {
+        toast({
+          title: `Service '${serviceName}' was not added`,
+          description: `Unable to connect to server '${serverName}'. Please make sure this server is online.`,
+          variant: "destructive"
+        });
+        setLoading(false);
+
+        //Closes dialog if failed
+        onClose();
+        return;
+      }
+
       // Create service using the API
+      const serviceData = {
+        providerId: parseInt(serverId),
+        name: serviceName,
+        serviceType: "SYSTEMD" as const,
+        serviceStatus: "unknown" as const
+      };
+      console.log('Creating systemd service with data:', serviceData);
+
       const response = await providerApi.createService(serviceData);
 
       console.log('Create systemd service response:', response);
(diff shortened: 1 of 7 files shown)


apps/client/src/pages/Alerts.tsx
```diff
@@ -123,6 +146,13 @@ export default function Alerts() {
                       <TableCell>
                         {new Date(alert.startsAt).toLocaleString()}
                       </TableCell>
+                      <TableCell>
+                        {(() => {
```
tamirg: lets  extract this one into function like the rest of the functions here

apps/client/src/pages/Dashboard.tsx
```diff
@@ -17,6 +17,15 @@ import { Button } from "@/components/ui/button"
...
 
+// Extract serviceId from alert (new field or from "id" like "fp:serviceId")
+const getAlertServiceId = (a: Alert): number | undefined => {
+  const anyA = a as any;
+  if (typeof anyA.serviceId === 'number') return anyA.serviceId;
+
+  const parts = a.id.split(':');         // "fingerprint:123"
+  const n = Number(parts[1]);
+  return Number.isFinite(n) ? n : undefined;
+};
```
tamirg: looks like its duplicated code, it appears also on TVMode.tsx file, please maybe extract into shared utils file in src/utils and implement just once and use in both places (lets say it will be Alert.utils.ts file)

tamirg (changes requested):
Nice work, had couple of comments (covered only FE side)

apps/client/src/pages/Dashboard.tsx
Oleksii101: Sure, let me fix it

Oleksii101:
@tamirg  its done

tamirg (approved):
In terms of client side, Approved ✅ 
@yuvalmich @Sahar541998 please approve from your side that its fine

Please do not merge until BE is approved too

Issue #305, opened 2025-10-03
[Bug] Drag-and-Drop Accepts All Files Instead of Allowed File Types


Aman-agraw-35:
Title: Drag-and-Drop Accepts All Files Instead of Allowed File Types

The drag-and-drop component for uploading secret files (SSH key or kubeconfig) currently accepts all file types, which is a security concern. Only specific file types should be allowed.

->Go to the secret upload section.

->Drag and drop any file (e.g., .txt, .jpg, .pdf).

->Observe that the file is accepted, even though it is not a valid SSH key or kubeconfig file.

The drag-and-drop upload should only accept valid file types, for example:

==>SSH keys: .pem, .ppk, .key

==>Kubeconfig: .yaml, .yml

Screenshots / Logs
<img width="641" height="590" alt="Image" src="https://github.com/user-attachments/assets/1e815956-3306-48c0-bc85-507c0df732d4" />

<img width="292" height="43" alt="Image" src="https://github.com/user-attachments/assets/1a850625-c066-4825-8722-85db65458f8d" />

I will fix this issue for hactober-fest-2k25

Manishhhsys:
Hi @idanlodzki ,  

I’d like to work on this issue as part of Hacktoberfest. Could you please assign it to me so I can start contributing?  

Thank you!

Aman-agraw-35:
@Manishhhsys I am giving this issue to you as I have other commitments to make. @idanlodzki Please assign the issue to him.

idanlodzki:
done @Manishhhsys good luck!

PR #344, opened 2025-10-07
[FIX]: Prevent drag-and-drop from accepting invalid file types (#305)

Manishhhsys:
## Issue Reference

See #305

---

## What Was Changed

- Updated `handleFile` function to correctly validate uploaded file extensions for SSH keys and Kubeconfig files.
- Updated `<FileDropzone>` component `accept` prop to only allow `.pem, .ppk, .pub, .yaml, .yml`.

---

## Why Was It Changed

The drag-and-drop uploader was previously allowing all file types, which could result in invalid secret uploads.  
This change ensures only valid SSH and Kubeconfig files are accepted and improves user experience by showing clear error messages.

---

## Screenshots

**Before:** Invalid files were accepted without warning.  
497366276-1e815956-3306-48c0-bc85-507c0df732d4

**After:** Invalid files are rejected and show a red error message below the uploader.  
Screenshot 2025-10-07 233241

---

- Valid files tested: `.pem`, `.ppk`, `.pub`, `.yaml`, `.yml` → accepted  
- Invalid files tested: `.txt`, `.json`, `.docx` → correctly rejected  
- This change does not affect other parts of the application.

### DIFF
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -21,7 +21,8 @@ import { Dashboard } from "@/components/Dashboard";
 import { ThemeProvider } from "./components/ThemeProvider";
 import { AuthGuard } from "./components/AuthGuard";
 import ScrollToTopButton from "@/components/ScrollToTopButton";
-
+import { isViewer } from "./lib/auth";
+import { Navigate } from "react-router-dom";
 const queryClient = new QueryClient();
 
 const App: React.FC = () => {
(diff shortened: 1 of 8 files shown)


idanlodzki:
@Manishhhsys 
Nice work adding the validation logic!
One note — checking only the file extension isn’t the right approach here, since users shouldn’t need to rename files just to upload them.
Instead, try reading the file content (for example with a FileReader) and checking if it contains something like
-----BEGIN PRIVATE KEY----- for SSH, or apiVersion: / kind: Config for kubeconfig.
That would make the validation much more accurate and user-friendly.

Manishhhsys:
Got it! That makes sense — checking the file content with FileReader would definitely improve accuracy.
However, I think it might be more secure to handle such validation on the server side, since any client-side check can be easily bypassed (e.g., by intercepting and modifying the upload request before it reaches the server using tools like burpsuite).

For now, this extension-based check provides a basic level of validation, but I agree that moving content validation to the backend would make it more robust.

Would love to hear your thoughts on this approach, @idanlodzki

idanlodzki:
@Manishhhsys 
Hey, I agree that the validation should be handled on the server side, so I’d like to ask you to move it there.

I also think the current check you added is unnecessary — when generating an SSH key file, it’s often created without an extension, and the same goes for kubeconfig files. It doesn’t make sense to expect users to rename their files. That’s both incorrect and frustrating, and it could easily discourage new contributors from working with OpsiMate.

Let me know if you’d like to take this task.
Also, are you already in our Slack channel? I’d love to discuss it there.
And by the way — did you give OpsiMate a ⭐ on GitHub?

Issue #350, opened 2025-10-07
[Bug] User registration returns 500/Internal server error


AnthonyFinney:
**Describe the bug**
When creating the first admin user from /register, certain passwords (e.g., short/weak ones) cause the API POST /api/v1/users/register to return HTTP 500 Internal Server Error with { success: false, error: "Internal server error" }

**To Reproduce**
1. Run the project locally (pnpm run dev)
2. Navigate to http://localhost:8080/register
3. Fill the form:
     Email: admin@myadmin.com
     Full name: admin
     Password: a short/weak password (e.g., admin)

**Expected behavior**
"password must be at least N characters long" this should be the error message.
Frontend should display that message to the user
Backend should not return HTTP 500 Internal Server Error

sudhir512kj:
Hi @AnthonyFinney i would like to work on this issue.

idanlodzki:
@sudhir512kj 
Hey, thanks for wanting to take this issue!
However, the original author is already working on it.
I’d love for you to grab another one — there are plenty available!
And of course, please join our Slack and give the repo a star ⭐

PR #351, opened 2025-10-08
fix: registerHandler now send the correct err msg

AnthonyFinney:
## Issue Reference

[Bug] User registration returns 500/Internal server error #350

---

## What Was Changed

I have changed the parsing order now the data from the form will be parsed then the try/catch will run.

---

## Why Was It Changed

it was sending incorrect error message

---

## Screenshots

before
Screenshot 2025-10-07 150423

after
image


apps/server/src/api/v1/users/controller.ts
```diff
@@ -13,8 +13,12 @@ export class UsersController {
     constructor(private userBL: UserBL) {}
 
     registerHandler = async (req: Request, res: Response) => {
+        const parsed = RegisterSchema.safeParse(req.body);
+        if(!parsed.success) {
+            return res.status(400).json({ success: false, error: 'Validation error', details: parsed.error.issues });
```
yuvalmich: It should have been caught in the zod error below

Can you check and elaborate what is the issue there and instead of changing the parse maybe catch the relevant error instead of zod error if this is the case

yuvalmich (commented):
Please review my comment and add a screenshot showing that it works as expected

apps/server/src/api/v1/users/controller.ts
AnthonyFinney: yes it should have and here is the error when i just console.log it - 
@OpsiMate/server:dev: ZodError: [ 
@OpsiMate/server:dev: { 
@OpsiMate/server:dev: "code": "too_small", 
@OpsiMate/server:dev: "minimum": 6, 
@OpsiMate/server:dev: "type": "string", 
@OpsiMate/server:dev: "inclusive": true, 
@OpsiMate/server:dev: "exact": false, 
@OpsiMate/server:dev: "message": "String must contain at least 6 character(s)", 
@OpsiMate/server:dev: "path": [ 
@OpsiMate/server:dev: "password" 
@OpsiMate/server:dev: ] 
@OpsiMate/server:dev: } 
@OpsiMate/server:dev: ]
but this is not hitting the zod error check and parsing of the data was creating the problem so moving it outside did produce the intendent error 

from my understanding when doing instanceof to check for ZodError it is the problem so using the safeParse and moving it outside so it will check for zod error

yuvalmich: I think I found the issue, it seems that the except part below when performing res.status...
We don't use return so it keeps to the next lines until it gets to the 500 error 

Can you try add a return statement in each of the error types and see if it solves it?

Can you search if we have other places like this in our code?

yuvalmich: If you can run in debug mode it would be the best so you can understand exactly what happens

Also can you please add a test that verify this issue? We have a folder containing all our tests

AnthonyFinney: ok on it

apps/server/tests/isZodError.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  test('returns false for non-Error values', () => {
+    expect(isZodError('not an error')).toBe(false);
+    expect(isZodError({})).toBe(false);
+    expect(isZodError(null)).toBe(false);
+    expect(isZodError(undefined)).toBe(false);
+  });
+
+  test('returns true when constructor.name is "ZodError"', () => {
+    const err = new Error('fake error') as any;
```
yuvalmich: Can you change the test that yields true on zod error to be actual zod parse the fails?

It will be more accurate test because we want to ensure that we catch an actual zod error

Issue #374, opened 2025-10-09
[Test] Test – Add API test for delete provider endpoint


idanlodzki:
# Test – Add API test for delete provider endpoint

## Describe the task
Add an automated API test to validate the delete provider endpoint. This test should ensure that providers can be deleted safely with proper authorization and cleanup.

## Test scenarios to include
✅ Successfully delete provider with valid provider ID and proper permissions.  
❌ Attempt to delete non-existent provider (should return 404).  
🔒 Verify correct permissions for provider deletion.  
🧾 Validate response structure (status code, confirmation message).  

## Expected behavior
The API test should confirm that:
- Valid requests return 200 OK or 204 No Content.
- Non-existent provider ID returns 404 Not Found.
- Unauthorized requests return 401/403 as appropriate.
- Provider is properly removed from database.
- Related services and data are handled appropriately (cascade delete or orphan handling).
- Audit log records the provider deletion.

## Additional context
- Use the existing API test framework (e.g., Jest, Supertest, or equivalent).
- Add the test under the server tests directory (`/apps/server/tests/api/providers.test.ts` or similar).
- Test with valid and invalid provider IDs, and verify cleanup of related data.

**Endpoint:** `DELETE /api/v1/providers/:providerId`

PR #414, opened 2025-10-12
[FEAT]: Add tests for delete provider endpoint

its-samrajya:
Issue: https://github.com/OpsiMate/OpsiMate/issues/374

## Changes

- Implemented comprehensive API tests for DELETE /api/v1/providers/:providerId.

- Verified cascade deletion of associated services and proper removal from the database.

- Ensured audit logs are correctly recorded, including user and provider details.


apps/server/tests/delete-provider.test.ts
```diff
@@ -0,0 +1,221 @@
+import request, { SuperTest, Test } from "supertest";
```
yuvalmich, line 1: Notice that we already have provider.test.ts file 

Please add a single specific test for deletion there, try to maintain the same functionality

idanlodzki: @cloudsamrajya  fixed?

PR #416, opened 2025-10-12
[FEAT]: Add delete provider test in provider.test.ts

its-samrajya:
## Issue Reference 
https://github.com/OpsiMate/OpsiMate/issues/374

---
## Description
This PR updates the delete provider test as per reviewer feedback on PR : https://github.com/OpsiMate/OpsiMate/pull/414

---

## What Was Changed

- Integrated deletion tests into existing provider.test.ts instead of creating a separate test file,
- Refactored admin user creation to run once in parent beforeEach hook instead of duplicating across each describe block

---

## Why Was It Changed

To reduce redundant code &  keep all provider-related tests together.

### DIFF
--- a/.github/workflows/sync-version
+++ b/.github/workflows/sync-version
@@ -1,0 +1,46 @@
+name: Sync Version on Release
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


yuvalmich:
Closing due to a big refactor in backend tests

Issue #404, opened 2025-10-11
[Feature] Add password reset via email


idanlodzki:
Is your feature request related to a problem? Please describe.
Currently, users who forget their password have no way to reset it through the UI. This creates a poor user experience and requires manual admin intervention.

Describe the solution you'd like
Implement a password reset via email feature that allows users to:

Click “Forgot password?” on the login page.

Receive an email with a secure, time-limited reset link.

Set a new password via a reset form.

Describe alternatives you’ve considered

Manual password reset by admins (not scalable or user-friendly).

Using external authentication providers (unnecessary for internal users).

Additional context

Will require the Mail Server Configuration feature (SMTP setup).

Reset links should expire after a short time (e.g., 15–30 minutes).

Log reset actions in the Audit Log for security visibility.

Ensure secure token generation and encrypted storage.

PR #454, opened 2025-10-13
[FEAT]: Add reset password by email

raphaeltx:
## Issue Reference

#404 

---

### Password Reset Table and Repository:
- Added a password_resets table and a PasswordResetsRepository class to securely store, retrieve, and manage password reset tokens.

### Forgot Password Logic:

- Generate a secure random token;
- Hash and store the token (encrypted) with an expiration time in the database;
- Send a password reset email to the user with a reset link containing the token.

### Reset Password Logic:

- Validate the token by checking its hash (after decryptation) and expiration in the database;
- Allow the user to set a new password if the token is valid;
- Check if the new password is not the same as the current password;
- Delete the token after successful password reset to prevent reuse;
- Log reset password actions in Audit Log table.

### Token Validation Endpoint:
- Created an endpoint to validate the reset token when the user loads the reset password page, ensuring expired or invalid tokens are handled smoothly.

### Success and Expired Pages (Frontend):
- Added components/pages to:

- Show a success message after a successful password reset.
- Display an error/expired message if the token is invalid or expired, with options to request a new reset link or return to login.

### MailService Implementation:

- Added a MailService class to encapsulate all email-sending logic using Nodemailer;
- I've create a 'service' layer to add externals integrations (we might need to define a better name);
- SMTP configuration is loaded from environment variables SMTP_HOST, SMTP_FROM, SMTP_PORT, SMTP_USER, SMTP_PASS;
- The transporter is initialized once in the constructor and verified on startup;
- The sendMail method provides a simple interface to send emails with subject, HTML, and recipient;
- All errors are logged and rethrown for better observability and error handling;
- The default sender address is set to "OpsiMate" <no-reply@opsimate.com>, but actual delivery depends on the SMTP provider’s configuration;

## **Important**

Requirements to run in production:
- Set up a smtp provider account;
- Create the secrets SMTP_USER and SMTP_PASS under the repository. These env variables are define under ci.yaml and the config.ts will replace the values in the configurations used by MailService.
- Set the provider account infos under default-config.ts

### Email Template added:
- Added the password reset email template to.
---

## Why Was It Changed

---

## Screenshots

Video Demo:

https://github.com/user-attachments/assets/ef837335-1cb9-40b8-990b-f779ed658102

Here's how the email looks on an actual email:

forgot_password_4

---

## Summary by CodeRabbit

* **New Features**
  * Added Forgot Password and Reset Password pages, forms, success and invalid-token screens; routes integrated into auth flow.
  * Users can request password resets and complete them via emailed secure links; server now exposes public password-reset endpoints and manages token lifecycle.
  * Mail delivery support: email templates, mail client, and mailing initialization.

* **Chores**
  * Mailer configuration defaults and base URL added/updated.

### DIFF
--- a/apps/server/src/bl/users/user.bl.ts
+++ b/apps/server/src/bl/users/user.bl.ts
@@ -125,7 +125,7 @@ export class UserBL {
         } catch (error) {
             logger.error('Failed to send password reset email', error);
             await this.passwordResetsRepo.deletePasswordResetsByUserId(user.id);
-            return;
+            throw new Error('Failed to send password reset email. Please try again later.');
         }
     }
 


raphaeltx:
Hi @idanlodzki, just pushed the changes

Looks good, if you have any suggestions just let me know

raphaeltx:
> > Hi @idanlodzki, just pushed the changes
> > Looks good, if you have any suggestions just let me know
> 
> Hey buddy would you share a demo of the same, maybe a video?

Hey @KaranNegi20Feb, definitely

I attached the video to the PR description

idanlodzki:
@raphaeltx almost there - take a look on the lint output

raphaeltx:
@idanlodzki done!

apps/client/src/components/ResetPasswordByEmailForm.tsx
```diff
@@ -0,0 +1,65 @@
+import { ErrorAlert } from "./ErrorAlert";
+import { Button } from "./ui/button";
+import { Input } from "./ui/input";
+
+interface ResetPasswordFormProps {
```
tamirg: suggestion - always keep callbacks at the bottom
helps in terms of readability in my opinion, input first output last

apps/client/src/components/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,46 @@
+import { ErrorAlert } from "./ErrorAlert";
+import { Button } from "./ui/button";
+import { Input } from "./ui/input";
+
+interface ForgotPasswordFormProps {
+  email: string;
```
tamirg: suggestion - always keep callbacks at the bottom
helps in terms of readability in my opinion, input first output last

default-config.yml
```diff
@@ -24,7 +24,7 @@ mailer:
   from: "<no-reply@yourdomain.com>"
   replyTo: "ops@yourdomain.com"
   auth:
-    user: "<Gmail username>"
-    pass: "<Gmail password>"        
+    user: ${SMTP_USER}
+    pass: ${SMTP_PASS}
```
yuvalmich: why is it needed?

.github/workflows/ci.yaml
```diff
@@ -39,6 +39,9 @@ jobs:
 
       - name: Build project
         run: pnpm run build
+        env:
+          SMTP_USER: ${{ secrets.SMTP_USER }}
+          SMTP_PASS: ${{ secrets.SMTP_PASS }}
```
yuvalmich: why is it needed?

apps/server/src/dal/external-client/mail-client.ts
```diff
@@ -0,0 +1,90 @@
+import { Logger } from "@OpsiMate/shared";
```
yuvalmich, line 1: I would rather to put the mail service implementation inside `dal/external-client/mail-client.ts`

same like `dal/external-client/datadog-client.ts` for example

apps/server/src/utils/mailTemplate.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Template for password reset email
+ * @param resetUrl
+ * @param userName
+ * @returns
+ */
+export function passwordResetTemplate(resetUrl: string, userName?: string) {
```
yuvalmich, line 7: this could be part of the MailClient class in my opinion (related to my earlier comment regarding changing it to an external client over a service)

apps/server/src/dal/passwordResetsRepository.ts
```diff
@@ -0,0 +1,92 @@
...
+            resetPassword.userId,
+            resetPassword.tokenHash,
+            resetPassword.expiresAt.toISOString()
+          );
+      });
+      transaction();
+    });
+  }
+
+  async getPasswordResetByTokenHash(
```
yuvalmich: I would really appreciate if you can keep the types standards of the DB table and the outer type the same like you used in `userRepository.ts` in `UserRow ` and `toSharedUser `

apps/server/src/config/config.ts
```diff
@@ -38,6 +38,29 @@ export interface OpsimateConfig {
 
 let cachedConfig: OpsimateConfig | null = null;
 
+/**
+ * Recursively replaces ${VAR} placeholders in all string fields of an object
+ * with the corresponding process.env.VAR value.
+ */
+function resolveEnvPlaceholders<T>(input: T): T {
```
yuvalmich: why is it necessary? why can't we keep using the same mechanism as the other config variables

yuvalmich (commented):
That's a really high quality contribution. I really appreciate your time and effort and the outcome is Great!

Left some comments, let me know if it make sense to you and if you need further assistance or information

.github/workflows/ci.yaml
raphaeltx: This one is for the user and password of the smtp provider. Without this we would have to commit these credentials under the default-config.ts file to make it work with the current config mechanism (not good keeping credentials exposed on github)

apps/server/src/config/config.ts
raphaeltx: Because of the credentials exposed issue, and also because yaml files do not support env vars replacement natively. if we do:

`
auth:
    user: ${SMTP_USER}
    pass: ${SMTP_PASS}
`

yaml file wont be able to read the values from env vars

raphaeltx:
@yuvalmich thanks man, i answered some of your comments and will push in a minute the other fixes

apps/server/src/utils/mailTemplate.ts
raphaeltx: About this one, MailClient is prepare to send different kinds of emails (not only the reset password), by passing the html in the sendMail function. Moving to the MailClient might grow the implementation in the future if more email templates appear. Do you think its a good approach?

apps/server/src/dal/external-client/mail-client.ts
raphaeltx: Done!

apps/server/src/dal/passwordResetsRepository.ts
raphaeltx: Done!

.github/workflows/ci.yaml
yuvalmich: But why do we need to add it in the `ci.yaml`
I'm probably missing something. @idanlodzki wdyt?

apps/server/src/bl/users/user.bl.ts
```diff
@@ -128,7 +128,7 @@ export class UserBL {
         } catch (error) {
             logger.error('Failed to send password reset email', error);
             await this.passwordResetsRepo.deletePasswordResetsByUserId(user.id);
-            throw new Error('Failed to send password reset email. Please try again later.');
+            return;
```
yuvalmich: why this error changed to an empty return statement?

.github/workflows/ci.yaml
idanlodzki: Hey, I see there’s no need for this. When a user wants to add an email service, they can include it in the configuration.
Also, I think by default the configuration can stay empty, with `enable` set to `false`.

default-config.yml
idanlodzki: the default value should be false

default-config.yml
idanlodzki: lets have some placeholders over here

user: "my-mail@gmail.com"
pass: "mypass"

for now

apps/server/src/bl/users/user.bl.ts
raphaeltx: this was a suggestion from coderabbit to avoid weird error messages to the user, now the message will be the success msg and mailer configuration instructions

KaranNegi20Feb:
Loved this PR @raphaeltx amazing work 😍.

PR #455, opened 2025-10-13
[FIX]: Implement password validation to prevent spaces in user passwords

### LINKED ISSUE #406: [Bug]  Users can create passwords containing spaces via admin panel
Describe the bug
When creating or updating a user account, the system currently allows passwords that contain spaces. This should not be permitted, as it can lead to inconsistent authentication behavior and potential validation issues.

To Reproduce

Go to User creation or password update form.

Enter a password that includes spaces (e.g., "My pass word123").

Observe that the password is accepted and the user is created successfully.

Expected behavior
Passwords should be rejected if they contain spaces. The validation should prompt the user with a clear message such as:

“Passwords cannot contain spaces.”

Additional context

Should apply both to user creation and password reset flows.

Add validation on both frontend and backend to ensure consistency.

Consider defining a clear password policy (min length, allowed characters).

HydrallHarsh:
Closes #406 
---

## What Was Changed
Implemented comprehensive password validation to reject passwords containing spaces across all user authentication flows, with consistent error messaging on both frontend and backend.

User Registration `Register.tsx` - Added space validation before API call
Admin User Creation  `AddUserModal.tsx` - Added JavaScript validation + HTML pattern
Admin Password Reset `ResetPasswordModal.tsx` - Added space validation
Profile Password Change `Profile.tsx` - Added space validation
Validation Feedback `ValidationFeedback.jsx` - Added password space rule for UI consistency
Schema Changes in `schemas.ts` and `controller.js`

---

### DIFF
--- a/apps/server/src/api/v1/users/controller.ts
+++ b/apps/server/src/api/v1/users/controller.ts
@@ -197,13 +197,6 @@ export class UsersController {
                     error: 'Password must be at least 8 characters long'
                 });
             }
-
-            if (/\s/.test(newPassword)) {
-                return res.status(400).json({
-                    success: false,
-                    error: 'Password must not contain spaces'
-                });
-            }
             
             // Don't allow admin to reset their own password this way
             if (userId === req.user.id) {


apps/server/src/api/v1/users/controller.ts
```diff
@@ -198,6 +198,13 @@ export class UsersController {
                 });
             }
 
+            if (/\s/.test(newPassword)) {
```
yuvalmich: It will never be executed because you defined the validation in the type itself

idanlodzki: @HydrallHarsh please resolve the conflicts and try fix this issue aswell

yuvalmich: Also please add a photo demonstrating the custom error message in the front end

Issue #340, opened 2025-10-06
[Feature] Add pagination to Dashboard services table


Oleksii101:
**Is your feature request related to a problem? Please describe.**
When there are many services (e.g., 30–100+), the Dashboard page becomes inconvenient to navigate — users have to scroll through a long list of servers.

**Describe the solution you'd like**
Implement pagination for the services table on the Dashboard page. 

<img width="1501" height="853" alt="Image" src="https://github.com/user-attachments/assets/92be8c6e-ca48-42e6-93c2-b804c436529b" />

Each page should display a fixed number of services (for example, 30 per page) with navigation controls (Next, Previous, Page numbers).

Users should also be able to select how many services to display per page — for example:
-10 per page

-30 per page

-50 per page

-100 per page
This option can be added as a dropdown or selector component above or below the table.

**Describe alternatives you've considered**

**Additional context**
Pagination should work smoothly with existing filters, search, and sorting.
The selected page size and current page number should ideally persist across page reloads.

idanlodzki:
@programmer765 any news?

programmer765:
@idanlodzki Still working on it. Had some health issues for last 2 days so couldn't give much time on it.

PR #468, opened 2025-10-16
[FEAT]: Add pagination feature to Dashboard for services table

programmer765:
## Issue Reference

- Closes [[Feature] Add pagination to Dashboard services table #340](https://github.com/OpsiMate/OpsiMate/issues/340)

---

## What Was Changed

- Made a component for pagination which handles the pagination logic for the services: `DashboardPagination.tsx`
- Made a component for handling the number of items to be viewed in a single page: `ItemsPerPageSelector.tsx`
- Added the number of pages calculation logic and services to display in the file `Dashboard.tsx`:

---

## Why Was It Changed

These changes makes scrolling through all the services a lot easier as per needed in issue #340 
---

## Screenshots

image
image
image

---

Currently all the `Items to be displayed per page` is hard coded in the file `ItemsPerPageSelector.tsx`. Do tell me if there's any kind of specific requirement for it.

### DIFF
--- a/apps/client/src/components/Dashboard/Dashboard.tsx
+++ b/apps/client/src/components/Dashboard/Dashboard.tsx
@@ -18,6 +18,7 @@ import { Button } from "@/components/ui/button"
 import { ChevronLeft, ChevronRight } from "lucide-react"
 import { cn } from "@/lib/utils"
 import { getAlertServiceId } from "@/utils/alert.utils";
+import DashboardPagination from "@/components/DashboardPagination"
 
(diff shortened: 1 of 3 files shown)


yuvalmich:
It won't work as expected with the current filters.

yuvalmich:
I wouldn't say inefficient on server side filtering where we have db indexing etc'

programmer765:
Got it. So should I implement it from server side? It would be much more complicated and time consuming i think

Issue #503, opened 2025-10-21
[Feature] Send welcome email after user creation


idanlodzki:
After a new user account is successfully created, send an automatic welcome email to the user.

Details:
	•	The email should include a short welcome message, project name, and useful links (e.g., documentation, Slack, GitHub).
	•	Trigger should occur right after user creation (e.g., after registration endpoint success).
	•	The email content should be customizable (subject and body).
	•	Consider using existing email configuration (SMTP settings) if available.

Example:
Subject: Welcome to OpsiMate!
Body:
Hey 👋
Welcome aboard! We’re happy to have you as part of the OpsiMate community.
You can start exploring here: [Docs link]
Join our Slack: [Slack link]

Acceptance Criteria:
	•	Email is sent automatically after registration
	•	Uses project’s email configuration
	•	Message is customizable via config or template file

amritamishra01:
Hi @idanlodzki,
I’m participating in Hacktoberfest 2025 and would like to work on this feature to send a welcome email after user registration. Looking forward to contributing!

idanlodzki:
Scope update after verifying against current main: self-registration (`registerUser`) already sends the welcome email. What's missing is the admin-creates-a-user path (`UserBL.createUser`) — the more important case, since that user never chose their own onboarding. This issue now covers only that path.

raphaeltx:
@idanlodzki hey man, long time :)

I just looked at this, do you want to discuss more? Feel free to reach out

PR #505, opened 2025-10-22
[FEAT] Add welcome email on user creation

raphaeltx:
## Issue Reference

#503 

---

## What Was Changed

* Added welcome email on user creation;
* Refactored Mail Client to handle new future emails;
* Added configuration options to customize subject and email body content;
* Refactored mail templates utility to reuse common email structures.

---

## Screenshots

opsimate_welcome_email

How does it look in a real email:
opsimate_welcome_email

---

## Summary by CodeRabbit

* **New Features**
  * Welcome emails are sent to newly registered admin users with personalized greetings.
  * Email configuration now supports a configurable welcome template (subject and body) and optional custom welcome content.
  * Email rendering uses a shared skeleton for consistent layout across messages.

* **Bug Fixes**
  * Email-sending failures during registration are caught and logged so registration still succeeds.

### DIFF
--- a/apps/server/src/dal/external-client/mail-client.ts
+++ b/apps/server/src/dal/external-client/mail-client.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@OpsiMate/shared';
 import nodemailer from 'nodemailer';
-import { getMailerConfig } from '../../config/config.js';
+import { getMailerConfig } from '../../config/config';
 import { passwordResetTemplate, welcomeTemplate } from '../../utils/mailTemplate';
 
 export enum MailType {


apps/server/src/dal/external-client/mail-client.ts
```diff
@@ -1,135 +1,176 @@
...
-    if (!token) {
-      throw new Error("Token is required for password reset email template");
-    }
+  /**
+   * Gets the welcome email template.
+   * @param options
+   * @returns string
+   */
+  private getWelcomeTemplate(options: SendMailOptions): string {
+    const body = options.customBody || this.mailerConfig?.options?.customBody;
```
yuvalmich: I would adjust the email options body to be more specific per template.

I mean I would change the config to be something like:
```

templates: {
    welcomeTemplate: {
        subject: 'xxxx',
        content: 'yyyyy'
    }
}
```

yuvalmich (commented):
Overall looks really good.

Left just one comment regarding supporting different email templates via the config

apps/server/src/dal/external-client/mail-client.ts
raphaeltx: Done!

apps/server/src/dal/external-client/mail-client.ts
```diff
@@ -1,125 +1,176 @@
 import { Logger } from '@OpsiMate/shared';
 import nodemailer from 'nodemailer';
-import { getMailerConfig } from '../../config/config';
-import { passwordResetTemplate } from '../../utils/mailTemplate';
+import { getMailerConfig } from '../../config/config.js';
```
yuvalmich: Notice that .js imports are no longer needed

apps/server/src/dal/external-client/mail-client.ts
```diff
@@ -1,125 +1,176 @@
...
+				return this.getResetPasswordTemplate(options);
+			case MailType.WELCOME:
+				return this.getWelcomeTemplate(options);
 			default:
 				throw new Error(`Unsupported mail type`);
 		}
 	}
 
+	private getMailSubject(options: SendMailOptions): string {
+		if (this.mailerConfig?.templates?.welcomeTemplate?.subject) {
```
yuvalmich: this should be inside the `case MailType.WELCOME`
we should return this if exists as you did, otherwise the default 'Welcome to OpsiMate!'

Can you please do the same for `options.subject` - should be passwordResetTemplate similarily to the welcome template

apps/server/src/dal/external-client/mail-client.ts
raphaeltx: Done!

apps/server/src/dal/external-client/mail-client.ts
raphaeltx: Done!

apps/server/src/dal/external-client/mail-client.ts
yuvalmich: You forgot one 🙂

yuvalmich (approved):

PR #519, opened 2025-10-25
[FEAT]: New mailer facility for general users 

KaranNegi20Feb:
The new PR will now keep the setup of forget password off when emailer config not fully configured.
### Changes:
1) New script and new way configure the **config.yml** and mount it to the docker environment so that when we change the **config.yml** (with the creds for emailer service) and bring in the new configs directly into container.
2) So i made a small api and used it to check the status until it's configured it won't show show "Forget Password?" when we finally configure it "Forget Password?" will appear.

A beautiful demo to whole procedure is here:

https://www.youtube.com/live/vdEEk6No0wc

## Summary by CodeRabbit

* **New Features**
  * Added email functionality support with optional password recovery feature. The "Forgot password" link now appears when email is enabled.
  * Enhanced email configuration options including sender address, reply-to address, and customizable email templates.

* **Chores**
  * Email features are now disabled by default and require proper SMTP configuration to activate.
  * Updated Docker deployment configuration for explicit config file handling.
  * Improved setup guidance for email feature enablement.


KaranNegi20Feb:
wow @yuvalmich really loved the new linter, didn't knew something this good came in nice 👍.

apps/server/src/api/health.ts
```diff
@@ -9,6 +13,22 @@ function healthCheck(req: Request, res: Response) {
 
 router.get('/health', healthCheck);
 
+router.get('/email-status', emailStatusHandler);
+
+function emailStatusHandler(req: Request, res: Response) {
+	try {
```
yuvalmich: is the try catch here is needed? I mean, is it possible that isEmailEnabled will raise an exception?

apps/server/src/api/v1/users/router.ts
```diff
@@ -20,6 +20,9 @@ export default function createUsersRouter(usersController: UsersController) {
 	// PATCH /users/profile - update user profile
 	router.patch('/profile', usersController.updateProfileHandler);
 
+	// GET /users/email-status - check if email functionality is enabled
+	router.get('/email-status', usersController.emailStatusHandler);
```
yuvalmich: why is it added twice? both here and in health.ts?

apps/server/src/api/v1/users/controller.ts
```diff
@@ -291,6 +298,13 @@ export class UsersController {
 				return res.status(400).json({ success: false, error: 'Validation error', details: error.errors });
 			}
 
+			if (error instanceof Error && error.message.includes('Email functionality is disabled')) {
```
yuvalmich: didn't you catch it already in line 287?

scripts/start-docker.sh
```diff
@@ -34,3 +34,13 @@ echo ""
 echo "OpsiMate is running!"
 echo "Frontend: http://localhost:8080"
 echo "Backend API: http://localhost:3001"
+echo ""
+echo "📧 Email Configuration:"
```
yuvalmich: I would remove all of these echo prints

KaranNegi20Feb: Umm for this when you run the script in the beginning there's no config for mailer, so instead this acts as an indicator that it's not set yet and you can set via the steps in echo.

So it's your call now do we keep it or should we redirect to the video? Or any better suggestion you have 😅😆

apps/server/src/api/v1/users/router.ts
KaranNegi20Feb: Yes you are right it shouldn't be like /user/email-status
It's actually just /email-status
The reason it's here in health is that it's the only place which has all the indicators like
- Health 
- / which says opsimate is running on ...

Well I'll just correct the comment in this
Or either create a new file like 
email-status and also correct the comment 😅

The reason why it's not in the 
V1/user/email-status is that why does it need a token and how will a guy have a token unless He's in or to be exact token is generated and can be used when he is logged in so 

Email status/email-status should be like /health, always present.

apps/server/src/api/v1/users/controller.ts
KaranNegi20Feb: Yeah pretty much 😅, I'll check it again

apps/server/src/api/health.ts
KaranNegi20Feb: Well it won't raise an exception though, like it's always available.
I am just habitual of putting everything in try and catch finally 😅.

I hope the linter does not catch me on this one.

apps/client/src/pages/Login.tsx
```diff
@@ -21,6 +22,27 @@ const Login: React.FC = () => {
...
+				} else {
+					logger.error('Failed to fetch email status', data);
+					setEmailEnabled(false);
+				}
+			} catch (err) {
+				logger.error('Failed to fetch email status', err);
+				setEmailEnabled(false);
+			}
+		};
+		fetchEmailStatus().catch((err) => {
```
tamirg: no need for catch as you already catching error inside the function

apps/client/src/pages/Login.tsx
```diff
@@ -21,6 +22,27 @@ const Login: React.FC = () => {
 		if (localStorage.getItem('jwt') && window.location.pathname === '/login') {
 			window.location.href = '/';
 		}
+		
+		// Fetch email enabled status from server
```
tamirg: remove comment

apps/client/src/pages/Login.tsx
KaranNegi20Feb: yeap agreed.

apps/client/src/pages/Login.tsx
KaranNegi20Feb: done, pushing.

scripts/start-docker.sh
KaranNegi20Feb: @yuvalmich i'll maybe shorten this or should i keep it this way? @idanlodzki @tamirg @Sahar541998 
for the time being we can keep it this way maybe, later we can decide and edit it,  it's small so not too much bothering i would say 😅

apps/server/src/api/v1/users/controller.ts
KaranNegi20Feb: Hey, yeah well. that's a different story like see http://localhost:3001/api/v1/users/forgot-password api's like these work when we send bearer token right but you see when the person is logging in he does not have the token with him so instead we have 2 , one which is POST req the one you mentioned and one is mine which is a GET req just for status.
I'll mark it as resolved @yuvalmich 😅

apps/server/src/api/v1/users/router.ts
KaranNegi20Feb: @yuvalmich on this one you are right i forgot to remove this one, it's redundant. Thankyou buddy 😅, resolved.

yuvalmich: I agree we can have just one route similar to the health one.
We also don't need authentication for this request

KaranNegi20Feb: yeap i have removed it, ill make a final pr in 5 mins

apps/server/src/api/health.ts
KaranNegi20Feb: yeap done this aswell now, no try catch since its a bool variable what possibly will it give other than true false so yes, thanks @yuvalmich 
 resolved.

apps/server/src/api/health.ts
```diff
@@ -1,14 +1,26 @@
...
 
 function healthCheck(req: Request, res: Response) {
 	res.send('ok');
 }
 
+function emailStatusHandler(req: Request, res: Response) {
+	const emailEnabled = isEmailEnabled();
+	return res.status(200).json({
+		success: true,
+		emailEnabled,
```
yuvalmich: just to keep the same standard like everywhere else please add emailEnabled into `data`

so it will be `data: { 'isEmailEnabled: true/false' }`

apps/server/src/api/health.ts
```diff
@@ -1,14 +1,26 @@
...
+		message: emailEnabled ? 'Email functionality is enabled' : 'Email functionality is disabled',
```
yuvalmich: I think we can remove the message, true/false is good enough. If we want to display a message in the client side we can determine it there

apps/server/src/api/v1/users/controller.ts
yuvalmich: but if the email functionality is disabled we will return 400 here:

`			if (!isEmailEnabled()) {
				return res.status(400).json({
					success: false,
					error: 'Email functionality is disabled. Please contact your administrator to configure SMTP settings.',
				});
			}`

apps/server/src/api/v1/users/controller.ts
```diff
@@ -337,4 +351,18 @@ export class UsersController {
 			}
 		}
 	};
+
+	emailStatusHandler = (req: Request, res: Response) => {
```
yuvalmich: this can be removed I believe. It as already implemented in the health router

scripts/start-docker.sh
yuvalmich: The issue is that even after the user will configure it, it will keep seeing these message which can be quite annoying.

Let's go with documentation on this one over the docker setup script

apps/server/src/api/v1/users/controller.ts
KaranNegi20Feb: done

apps/server/src/api/health.ts
KaranNegi20Feb: {
  "success": true,
  "data": {
    "isEmailEnabled": true
  }
}
here you go, it's now like this.

idanlodzki:
@KaranNegi20Feb  sup king? just resolve the last conflict and i will merge

PR #539, opened 2025-10-31
[FEAT] Add custom actions feature with CRUD support

Sahar541998:
## Summary
This PR adds a new custom actions feature with full CRUD support on the server side.

## Changes
- **New shared package**: `@OpsiMate/custom-actions` with TypeScript types for actions
- **CRUD API**: Full create, read, update, delete endpoints for custom actions
- **Action execution**: Support for running bash actions on providers and services
- **Provider integration**: Extended `ProviderConnector` interface to support custom actions
- **Database schema**: New `custom_actions` table with support for bash scripts
- **Tests**: Comprehensive API tests for custom actions
- **Docker integration**: Updated all Dockerfiles to include the custom-actions package

## Implementation Details
- Bash actions are fully implemented and can be executed via SSH
- HTTP action types are defined in the shared package but not yet implemented
- Repository returns `CustomAction` types from `@OpsiMate/custom-actions`
- Action execution supports both provider-level and service-level targets

## Testing
- All CRUD operations tested via API endpoints
- Action execution tested for both provider and service targets
- Linting passes with no errors

## Summary by CodeRabbit

* **New Features**
  * Full Custom Actions: create, edit, delete, and run Bash or HTTP actions (with headers, method, body).
  * New Actions area: categorized action cards, two‑step modal form, action list with empty‑state and create flow.
  * Run actions from dashboard: inline selector, dedicated Run dialog to target a service or provider, and running state feedback.
  * Dashboard & navigation: run-action controls integrated into service action bar and new "Actions" sidebar item.

### DIFF
--- a/apps/server/tests/setup.ts
+++ b/apps/server/tests/setup.ts
@@ -28,6 +28,29 @@ vi.mock('@kubernetes/client-node', () => ({
 	NetworkingV1Api: vi.fn(),
 }));
 
+// Mock the SSH client to avoid actual SSH connections in tests
+vi.mock('../src/dal/sshClient', async () => {
+	const actual = await vi.importActual('../src/dal/sshClient');
+	return {
+		...actual,
+		executeBashScript: vi.fn().mockResolvedValue({
+			code: 0,
+			stdout: 'mocked output',
+			stderr: '',
+			signal: null,
+		}),
+		testConnection: vi.fn().mockResolvedValue({ success: true }),
+		connectAndListContainers: vi.fn().mockResolvedValue([]),
+		startService: vi.fn().mockResolvedValue(undefined),
+		stopService: vi.fn().mockResolvedValue(undefined),
+		getServiceLogs: vi.fn().mockResolvedValue(['mocked log']),
+		startSystemService: vi.fn().mockResolvedValue(undefined),
+		stopSystemService: vi.fn().mockResolvedValue(undefined),
+		getSystemServiceLogs: vi.fn().mockResolvedValue(['mocked system log']),
+		checkSystemServiceStatus: vi.fn().mockResolvedValue('running'),
+	};
+});
+
 // Increase timeout for integration tests
 vi.setConfig({ testTimeout: 30000 });
 


packages/custom-actions/eslint.config.mjs
```diff
@@ -0,0 +1,25 @@
...
+			parserOptions: {
+				project: './tsconfig.json',
+			},
+		},
+		rules: {
+			'@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_' }],
+			'@typescript-eslint/no-explicit-any': 'warn',
+			'@typescript-eslint/no-floating-promises': 'error',
+			'no-console': 'warn',
+		},
```
tamirg: Is there a reason we need them specifically for this package?

apps/server/src/dal/customActionRepository.ts
```diff
@@ -0,0 +1,130 @@
...
+type CustomActionRow = {
+	id: number;
+	name: string;
+	description: string;
+	type: 'bash' | 'http';
+	target: 'service' | 'provider' | null;
+	script: string | null;
+	http_config: string | null;
+	created_at: string;
+};
```
tamirg: Can we use enums and types you already defined in the package in some of these fields where ever possible?

packages/custom-actions/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
tamirg, line 1: @yuvalmich we might want to extend base tsconfig as well for all packages/apps

turbo.json
```diff
@@ -51,7 +51,9 @@
         "^build"
       ],
       "outputs": [],
-      "env": ["NODE_ENV"]
+      "env": [
```
tamirg, line 54: Is it a format thing? If so how did it not get formatted already?

apps/server/src/bl/providers/provider-connector/vmProviderConnector.ts
```diff
@@ -1,44 +1,63 @@
-import { DiscoveredPod, DiscoveredService, Provider, Service, ServiceType } from '@OpsiMate/shared';
+import {DiscoveredPod, DiscoveredService, Provider, Service, ServiceType, Logger} from '@OpsiMate/shared';
```
tamirg: What is happening here? I think this is not formatted

packages/custom-actions/eslint.config.mjs
Sahar541998: No
Nice catch

apps/server/src/dal/customActionRepository.ts
Sahar541998: I  have separate type for the row in the db and the actial type we will pass in the server to the client

tamirg: Yes but consider sharing at least the target and type fields as they have pretty much same contract 
it will be easier to refactor/extend in the future without needing to handle both places

packages/custom-actions/tsconfig.json
yuvalmich: Agree

apps/server/src/api/v1/custom-actions/router.ts
```diff
@@ -0,0 +1,22 @@
...
+import PromiseRouter from 'express-promise-router';
+import { CustomActionsController } from './controller';
+
+export default function createCustomActionsRouter(controller: CustomActionsController) {
+	const router = PromiseRouter();
+
+	// CRUD
+	router.get('/', controller.list);
+	router.post('/', controller.create);
+	router.get('/:actionId', controller.getById);
```
yuvalmich, line 11: We need to notice if the client uses this endpoint or we can remove it

apps/server/src/api/v1/custom-actions/controller.ts
```diff
@@ -0,0 +1,101 @@
...
+import {CustomActionRepository} from '../../../dal/customActionRepository';
+import {providerConnectorFactory} from '../../../bl/providers/provider-connector/providerConnectorFactory';
+import {CustomAction} from '@OpsiMate/custom-actions';
+
+const logger: Logger = new Logger('api/custom-actions');
+
+export class CustomActionsController {
+    constructor(
+        private bl: CustomActionBL,
+        private providerRepo: ProviderRepository,
```
yuvalmich: Let's try to work here only with BL level. More than that, let's try to get here only the  CustomActionBL and inside the BL to use only other BLs (and its own repository)

apps/server/src/api/v1/custom-actions/controller.ts
```diff
@@ -0,0 +1,101 @@
...
+        private serviceRepo: ServiceRepository,
+        private actionRepo: CustomActionRepository
+    ) {
+    }
+
+    private isValidBashAction(body: unknown): body is CustomAction {
```
yuvalmich: Can we achieve this using zod?

apps/server/src/api/v1/custom-actions/controller.ts
```diff
@@ -0,0 +1,101 @@
...
+        const action = await this.actionRepo.getById(actionId);
+        if (!action) return res.status(404).json({success: false, error: 'Action not found'});
+
+        const connector = providerConnectorFactory(provider.providerType);
+        await connector.runCustomAction(provider, action);
+        logger.info(`Ran custom action id=${actionId} on provider id=${providerId}`);
+        return res.status(200).json({success: true});
+    };
+
+    runForService = async (req: Request, res: Response) => {
```
yuvalmich: I guess the run needs more work, it can get either service or provider and not always both

apps/server/src/api/v1/custom-actions/controller.ts
```diff
@@ -0,0 +1,101 @@
...
+        const actionId = Number(req.params.actionId);
+        const service = await this.serviceRepo.getServiceById(serviceId);
+        if (!service) return res.status(404).json({success: false, error: 'Service not found'});
+        const provider = await this.providerRepo.getProviderById(service.providerId);
+        if (!provider) return res.status(404).json({success: false, error: 'Provider not found'});
+        const action = await this.actionRepo.getById(actionId);
+        if (!action) return res.status(404).json({success: false, error: 'Action not found'});
+
+        const connector = providerConnectorFactory(provider.providerType);
+        await connector.runCustomAction(provider, action, service);
```
yuvalmich: this logic is true only for bash implementation. maybe we can make it a bit more generic to be able to add easily different actions.

I thought this function would be in the CustomActionBL layer. and it will implement some kind of generic factory like the providerConnector that will have a different implementation per action type (bash, http, jenkins, ...) and then this code will be in the bash implementation

apps/server/src/dal/customActionRepository.ts
yuvalmich: I agree, it is ok we have different general types between the transportation layer and storage layer but for the enums themself we want to share them from the shared package IMO

apps/server/src/api/v1/custom-actions/controller.ts
Sahar541998: If it run fot service we always have privider 
If its for provider then only provider and the third option is non of above

apps/server/src/dal/customActionRepository.ts
```diff
@@ -0,0 +1,130 @@
+import Database from 'better-sqlite3';
+import { runAsync } from './db';
+import { CustomAction } from '@OpsiMate/custom-actions';
+
+type CustomActionRow = {
```
yuvalmich: we have a models.ts file in which we store all of the db schemas. the purpose of it, is when different repositories use the same model (i.e. provider joins with service in the provider repo)

I think it will be better to move it to there (I know it a bit messy at the moment but I thought maybe refactoring it to a model folder where every file will hold a different repo schema) let me know what you think

apps/server/src/api/v1/custom-actions/controller.ts
yuvalmich: Right I missed it

apps/server/src/api/v1/custom-actions/router.ts
Sahar541998: @tamirg you need action by id command?

apps/server/src/api/v1/custom-actions/controller.ts
Sahar541998: i think the abstraction should be inside the ProviderConnector. which now dosent exist

it just throw 		return Promise.reject(new Error('Custom actions are not implemented for K8S'));

next step i will construct here something maybe builder? to make it support all types of actions 
not sure how yet.

tamirg (approved):

PR #540, opened 2025-10-31
[FEAT][Custom Actions] Create actions page UI

tamirg:
## What Was Changed

1. Added chakra-ui modal usage inside ActionModal in order to start and integrate new design system library.
2. Added new Actions page to allow showing actions for providers and services as well as allowing to add/edit them with a modal.

Action could be created with a service/provider targets and would have either bash type or http request.

---

## Why Was It Changed

---

## Screenshots

Actions page:
image

image

HTTP:
image

BASH:
image

---


tamirg:
@coderabbitai review

Issue #652, opened 2026-07-04
Audit log: creating mutepolicy is not recorded


idanlodzki:
## Problem

Creating (and updating/deleting) an **alert silence** leaves no trace in the audit log. Silences directly affect what the team sees firing, so knowing who silenced what — and when — matters. Other resources (providers, users, dashboards, secrets) already write audit entries.

`apps/server/src/bl/silences/silence.bl.ts` currently has no audit calls at all.

## How to fix

Follow the existing pattern in `apps/server/src/bl/providers/provider.bl.ts` (see `createProvider`, which calls `this.auditBL.logAction({...})` after the mutation succeeds):

1. Add `SILENCE` to the `AuditResourceType` enum in `packages/shared/src/types.ts` (the enum has an "Add more as needed" comment).
2. Inject `AuditBL` into `SilenceBL` (constructor DI, wired in `apps/server/src/app.ts`).
3. Call `auditBL.logAction` with `AuditActionType.CREATE` / `UPDATE` / `DELETE` after each mutation, passing the acting user's id/name from the authenticated request.

## Notes

- The audit write should be best-effort: a failed audit insert must not fail the mutation (see `recordHistoryEvent` in `alert.bl.ts` for the pattern).
- Requests authenticated with an API token have no `req.user`; handle the missing-actor case gracefully rather than crashing.

## Acceptance criteria

- Creating a silence produces an audit log entry visible in the Audit page (action type, resource type, silence name, acting user).
- Update and delete are covered too.
- A server test in `apps/server/tests/` asserting the audit row is written on create.

VitorBernalRodrigues:
Hi! I’d like to work on this.

I confirmed the audit gap still exists on `main` after the Silences → Mute Policies rename. I plan to follow the existing `EnrichmentBL` audit pattern for create/update/delete, including best-effort logging, the API-token actor fallback, and focused server regression coverage.

Since the current domain is now Mute Policy, I’d use `MUTE_POLICY` as the audit resource type unless you prefer preserving the older `SILENCE` naming.

Issue #651, opened 2026-07-04
Audit log: creating custom actions is not recorded


idanlodzki:
## Problem

Creating (and updating/deleting) a **custom action** leaves no trace in the audit log. Other resources already do this — e.g. providers, users, dashboards, secrets — so actions are a blind spot for anyone reviewing who changed what.

`apps/server/src/bl/actions/action.bl.ts` currently has no audit calls at all.

## How to fix

Follow the existing pattern in `apps/server/src/bl/providers/provider.bl.ts` (see `createProvider`, which calls `this.auditBL.logAction({...})` after the mutation succeeds):

1. Add `ACTION` to the `AuditResourceType` enum in `packages/shared/src/types.ts` (the enum has an "Add more as needed" comment).
2. Inject `AuditBL` into `ActionBL` (constructor DI, wired in `apps/server/src/app.ts`).
3. Call `auditBL.logAction` with `AuditActionType.CREATE` / `UPDATE` / `DELETE` after each mutation, passing the acting user's id/name from the authenticated request.

## Notes

- The audit write should be best-effort: a failed audit insert must not fail the mutation (see how alert history events are handled in `alert.bl.ts` → `recordHistoryEvent`).
- Requests authenticated with an API token have no `req.user`; handle the missing-actor case gracefully rather than crashing.

## Acceptance criteria

- Creating a custom action produces an audit log entry visible in the Audit page (action type, resource type, resource name, acting user).
- Update and delete are covered too.
- A server test in `apps/server/tests/` asserting the audit row is written on create.

PR #663, opened 2026-07-05
feat(server): audit logging for enrichment mutations

### LINKED ISSUE #653: Audit log: creating enrichment rules is not recorded
## Problem

Creating (and updating/deleting) an **enrichment rule** leaves no trace in the audit log. Enrichments rewrite alert content (summary templates, added fields) for everyone, so changes to them are worth auditing. Other resources (providers, users, dashboards, secrets) already write audit entries.

`apps/server/src/bl/enrichments/enrichment.bl.ts` currently has no audit calls at all. (Enrichments do stamp `last_modified_by` on the row itself, but that only keeps the latest editor — it is not a history.)

## How to fix

Follow the existing pattern in `apps/server/src/bl/providers/provider.bl.ts` (see `createProvider`, which calls `this.auditBL.logAction({...})` after the mutation succeeds):

1. Add `ENRICHMENT` to the `AuditResourceType` enum in `packages/shared/src/types.ts` (the enum has an "Add more as needed" comment).
2. Inject `AuditBL` into `EnrichmentBL` (constructor DI, wired in `apps/server/src/app.ts`).
3. Call `auditBL.logAction` with `AuditActionType.CREATE` / `UPDATE` / `DELETE` after each mutation, passing the acting user's id/name from the authenticated request (the controller already threads the actor for the `last_modified_by` stamp — reuse that).

## Notes

- The audit write should be best-effort: a failed audit insert must not fail the mutation (see `recordHistoryEvent` in `alert.bl.ts` for the pattern).
- Requests authenticated with an API token have no `req.user`; handle the missing-actor case gracefully rather than crashing.

## Acceptance criteria

- Creating an enrichment rule produces an audit log entry visible in the Audit page (action type, resource type, rule name, acting user).
- Update and delete are covered too.
- A server test in `apps/server/tests/` asserting the audit row is written on create.

hfl0506:
## Issue Reference
Closes #653 

---

## What Was Changed
- Added ENRICHMENT to AuditResourceType.
- Injected AuditBL into EnrichmentBL.
- Added audit logging after enrichment create, update, and delete mutations.
- Passed the authenticated req.user from enrichment controller methods into the BL.
- Added graceful fallback actor handling for API-token requests with no req.user.
- Made enrichment audit writes best-effort, so audit failures do not fail the mutation.
- Added a server test asserting enrichment creation writes anaudit row.

---

## Why Was It Changed
- Enrichment rules change alert content for all users, so create/update/delete actions need historical traceability.
- last_modified_by only stores the latest editor, not a full audit history.
- Existing resources already use audit logging, so enrichments now follow the same pattern.
- API-token authenticated requests should keep working even without a user object.
- Audit logging should not block the actual enrichment mutation if the audit insert fails.
- The new test prevents regressions for create audit behavior.

---

## Test Result
Ran `pnpm test`

### Step 1 - create enrichment
Screenshot 2026-07-05 at 9 41 47 AM

### Step 2 - update enrichment
Screenshot 2026-07-05 at 9 42 09 AM

### Step 3 - delete enrichment
Screenshot 2026-07-05 at 9 42 33 AM

## Summary by CodeRabbit

* **New Features**
  * Enrichment create, update, and delete actions now emit audit entries tagged to an **ENRICHMENT** resource type, including the acting user.
* **Bug Fixes**
  * Improved audit attribution to consistently use authenticated user context for enrichment actions, including accurate actor details.
  * Enrichment audit logs are recorded only when appropriate (e.g., no audit for empty update payloads).
* **Tests**
  * Added coverage for enrichment audit lifecycle (create/update/delete) and verified no audit entries are produced for empty update payloads.

### DIFF
--- a/apps/server/src/api/v1/enrichments/controller.ts
+++ b/apps/server/src/api/v1/enrichments/controller.ts
@@ -53,7 +53,7 @@ export class EnrichmentController {
 					summaryTemplate: data.summaryTemplate ?? null,
 					priority: data.priority ?? 0,
 				},
-				req.user?.fullName
+				req.user
 			);
 			return res.status(201).json({ success: true, data: enrichment, message: 'Enrichment created' });
 		} catch (error) {
@@ -75,7 +75,7 @@ export class EnrichmentController {
 				return res.status(404).json({ success: false, error: 'Enrichment not found' });
 			}
 
-			const updated = await this.enrichmentBL.update(enrichmentId, data, req.user?.fullName);
+			const updated = await this.enrichmentBL.update(enrichmentId, data, req.user);
 			return res.json({ success: true, data: updated, message: 'Enrichment updated' });
 		} catch (error) {
 			if (isZodError(error)) {
@@ -86,14 +86,14 @@ export class EnrichmentController {
 		}
 	};
 
-	deleteHandler = async (req: Request, res: Response) => {
+	deleteHandler = async (req: AuthenticatedRequest, res: Response) => {
 		try {
 			const { enrichmentId } = AlertEnrichmentIdSchema.parse({ enrichmentId: req.params.enrichmentId });
 			const existing = await this.enrichmentBL.get(enrichmentId);
 			if (!existing) {
 				return res.status(404).json({ success: false, error: 'Enrichment not found' });
 			}
-			await this.enrichmentBL.delete(enrichmentId);
+			await this.enrichmentBL.delete(enrichmentId, req.user);
 			return res.json({ success: true, message: 'Enrichment deleted' });
 		} catch (error) {
 			if (isZodError(error)) {
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -171,7 +171,7 @@ export async function createApp(db: Database.Database, mode: AppMode): Promise<e
 	const playgroundBL = new PlaygroundBL(playgroundRepo, mailClient);
 	const silenceBL = new SilenceBL(silenceRepo);
 	alertBL.setSilenceBL(silenceBL);
-	const enrichmentBL = new EnrichmentBL(enrichmentRepo);
+	const enrichmentBL = new EnrichmentBL(enrichmentRepo, auditBL);
 	alertBL.setEnrichmentBL(enrichmentBL);
 	const actionBL = new ActionBL(actionRepo, alertHistoryRepo);
 
--- a/apps/server/src/bl/enrichments/enrichment.bl.ts
+++ b/apps/server/src/bl/enrichments/enrichment.bl.ts
@@ -40,9 +40,10 @@ export class EnrichmentBL {
 	}
 
 	async update(id: number, data: UpdateEnrichmentInput, actor?: User | null): Promise<AlertEnrichment | undefined> {
+		const shouldAudit = this.hasUpdateData(data);
 		await this.enrichmentRepo.updateEnrichment(id, data, actor?.fullName);
 		const updated = await this.enrichmentRepo.getEnrichmentById(id);
-		if (updated) {
+		if (updated && shouldAudit) {
 			await this.recordAuditAction(AuditActionType.UPDATE, updated, actor, JSON.stringify(data));
 		}
 		return updated;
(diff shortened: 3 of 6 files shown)


idanlodzki (changes requested):
@hfl0506  First off — thank you for picking this up so quickly, and welcome! 🎉 This is a really solid take on #653: the audit wiring follows the provider pattern cleanly, you caught both gotchas from the issue (best-effort audit writes, API-token requests without `req.user`), and the extra touches — passing the full `User` instead of just a name, skipping audit entries for empty update payloads, capturing the row before delete so the entry has the real resource name — are exactly right. Tests cover the lifecycle well.

One change I need before merging, and two small optional notes:

**1. Please revert the `packages/shared/package.json` change (required)**

Pointing `types` at `./src/index.ts` instead of `./dist/index.d.ts` is unrelated to audit logging and changes how every consumer in the monorepo resolves the package's types. I suspect you hit the stale-`dist` pain (add an enum value, TS can't see it until `pnpm --filter @OpsiMate/shared build`) — totally real, we've hit it too. But source-pointing types trades a *visible* type error for a *silent* runtime one: your code typechecks against `src` while the runtime still loads the stale `dist`, which crashes at import time (we debugged exactly that failure mode in the client recently). If you think the DX tradeoff is worth it, I'd genuinely welcome that as its own small PR/issue where we can discuss it properly — it just shouldn't ride along here.

**2. Title nit (optional)**

`feat(server): add enrichment mutations` undersells/misdescribes it — the mutations already existed. Something like `feat(server): audit logging for enrichment mutations` reads better in the changelog.

**3. For the follow-ups (no action needed)**

`getAuditActor` + `recordAuditAction` are nicely generic — #651 (actions) and #652 (silences) will want the identical logic. If you feel like extracting them somewhere reusable (or picking up one of those issues too 😉), that'd be fantastic, but it's fine to leave that for the next PR.

Revert the package.json change and this is good to go. Thanks again!

hfl0506:
> @hfl0506 First off — thank you for picking this up so quickly, and welcome! 🎉 This is a really solid take on #653: the audit wiring follows the provider pattern cleanly, you caught both gotchas from the issue (best-effort audit writes, API-token requests without `req.user`), and the extra touches — passing the full `User` instead of just a name, skipping audit entries for empty update payloads, capturing the row before delete so the entry has the real resource name — are exactly right. Tests cover the lifecycle well.
> 
> One change I need before merging, and two small optional notes:
> 
> **1. Please revert the `packages/shared/package.json` change (required)**
> 
> Pointing `types` at `./src/index.ts` instead of `./dist/index.d.ts` is unrelated to audit logging and changes how every consumer in the monorepo resolves the package's types. I suspect you hit the stale-`dist` pain (add an enum value, TS can't see it until `pnpm --filter @OpsiMate/shared build`) — totally real, we've hit it too. But source-pointing types trades a _visible_ type error for a _silent_ runtime one: your code typechecks against `src` while the runtime still loads the stale `dist`, which crashes at import time (we debugged exactly that failure mode in the client recently). If you think the DX tradeoff is worth it, I'd genuinely welcome that as its own small PR/issue where we can discuss it properly — it just shouldn't ride along here.
> 
> **2. Title nit (optional)**
> 
> `feat(server): add enrichment mutations` undersells/misdescribes it — the mutations already existed. Something like `feat(server): audit logging for enrichment mutations` reads better in the changelog.
> 
> **3. For the follow-ups (no action needed)**
> 
> `getAuditActor` + `recordAuditAction` are nicely generic — #651 (actions) and #652 (silences) will want the identical logic. If you feel like extracting them somewhere reusable (or picking up one of those issues too 😉), that'd be fantastic, but it's fine to leave that for the next PR.
> 
> Revert the package.json change and this is good to go. Thanks again!

Thanks for review! 
Reverted the shared package types change. I’ll leave the audit helper extraction for a follow-up PR when working on #651/#652 so this PR stays focused.

idanlodzki (approved):
All addressed — the package.json revert, the title, and the CodeRabbit nits. Thanks for the quick turnaround and for a genuinely well-built first PR: the actor handling and best-effort audit pattern here are exactly what #651 and #652 should copy.

Approved ✅ — and the Slack door is open if you want to hang out with the community: https://join.slack.com/t/opsimate/shared_invite/zt-39bq3x6et-NrVCZzH7xuBGIXmOjJM7gA

PR #675, opened 2026-07-13
feat(server): audit logging for action mutations

### LINKED ISSUE #651: Audit log: creating custom actions is not recorded

mvanhorn:
## Issue Reference

Closes #651

---

## What Was Changed

Custom actions are now recorded in the audit log on create, update, and delete.

- `packages/shared/src/types.ts` — added `ACTION` to the `AuditResourceType` enum.
- `apps/server/src/bl/actions/action.bl.ts` — injected `AuditBL` into `ActionBL` and added best-effort `recordAuditAction` calls to `create`/`update`/`delete`, mirroring the existing `EnrichmentBL` implementation.
- `apps/server/src/api/v1/actions/controller.ts` — typed the create/update/delete handlers as `AuthenticatedRequest` and passed `req.user` through as the acting user (the run handler already did this).
- `apps/server/src/app.ts` — wired the existing `auditBL` into the `ActionBL` constructor.
- `apps/server/tests/action-audit.test.ts` — new server test asserting the audit row on create, plus create/update/delete coverage, following `audit.test.ts`.

---

## Why Was It Changed

Creating, updating, or deleting an action left no trace in the audit log, while every other mutable resource (providers, users, dashboards, secrets, enrichments) already writes audit rows. Actions were therefore a blind spot on the Audit page.

The fix mirrors the enrichment audit pattern (PR #663) exactly, so it stays consistent with the rest of the codebase:

- Audit writes are best-effort — wrapped in try/catch and logged on failure — so a failing audit insert never fails the mutation itself.
- API-token requests that carry no `req.user` do not crash; the actor falls back to id `0` / `API Token`.

---

## Screenshots

N/A - server-only change (audit-log record writes); no UI or visual diff. Behavior is covered by the new `action-audit` server test.

---

Verified with `pnpm --filter @OpsiMate/server test` (115 tests pass, including the new `action-audit` file) and a `tsc --noEmit` typecheck with no new errors.

### DIFF
--- a/apps/server/src/api/v1/actions/controller.ts
+++ b/apps/server/src/api/v1/actions/controller.ts
@@ -43,16 +43,19 @@ export class ActionController {
 		}
 	};
 
-	createHandler = async (req: Request, res: Response) => {
+	createHandler = async (req: AuthenticatedRequest, res: Response) => {
 		try {
 			const data = CreateActionSchema.parse(req.body);
-			const action = await this.actionBL.create({
-				name: data.name,
-				type: data.type,
-				config: data.config,
-				nameContains: data.nameContains ?? null,
-				labelMatchers: data.labelMatchers ?? [],
-			});
+			const action = await this.actionBL.create(
+				{
+					name: data.name,
+					type: data.type,
+					config: data.config,
+					nameContains: data.nameContains ?? null,
+					labelMatchers: data.labelMatchers ?? [],
+				},
+				req.user
+			);
 			return res.status(201).json({ success: true, data: action, message: 'Action created' });
 		} catch (error) {
 			if (isZodError(error)) {
@@ -115,7 +118,7 @@ export class ActionController {
 		}
 	};
 
-	updateHandler = async (req: Request, res: Response) => {
+	updateHandler = async (req: AuthenticatedRequest, res: Response) => {
 		try {
 			const { actionId } = ActionIdSchema.parse({ actionId: req.params.actionId });
 			const data = UpdateActionSchema.parse(req.body);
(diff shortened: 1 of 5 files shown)


idanlodzki:
Thanks for this @mvanhorn — really solid first contribution! 🎉

I reviewed the code and it's exactly how we'd want it done: it mirrors the enrichment audit pattern from #663 faithfully (best-effort audit writes that never fail the mutation, API-token fallback actor), the delete handler correctly snapshots the action *before* deleting so the audit row keeps its name, and the test coverage is great. I also double-checked the Audit page in the client — it renders `resourceType` as-is, so `ACTION` shows up there with no client changes needed. Ran the full server suite locally on your branch: 115/115 passing.

The only thing failing CI was Prettier formatting on `tests/action-audit.test.ts` (one line wrapped differently), so I pushed a tiny `style:` commit to your branch to get CI green rather than round-trip it back to you. For next time: `pnpm --filter @OpsiMate/server format-fix` before pushing will keep the formatter happy — CI runs `format` (check mode) on every package.

Looking forward to more contributions! 🚀

PR #731, opened 2026-08-02
fix: drag-select marking stops when cursor leaves checkbox column

### LINKED ISSUE #712: [Bug] Drag-select marking stops when the cursor leaves the checkbox column
**Describe the bug**
When marking alerts by dragging (mouse down on a checkbox, then moving down the list), the marking only works while the cursor stays inside the checkbox column. Moving the cursor slightly right — still over the same alert rows — stops rows from being marked, even though the drag continues.

**To Reproduce**
1. Go to the Alerts table.
2. Mouse-down on the first row's checkbox and drag downward, drifting horizontally over the row content.
3. Rows the cursor passes over outside the checkbox column are not selected.

**Expected behavior**
While a drag-selection is active, moving over any part of a row should mark it — the drag anchor already established the intent; the horizontal position shouldn't matter.

**Additional context**
Drag selection lives in `apps/client/src/components/Alerts/AlertsTable/hooks/useDragSelection.ts`; the `onDragEnter` handlers are wired per-row in `AlertRow.tsx` — likely attached to the checkbox cell only rather than the whole row while dragging.

waterWang:
## Summary

Fixes #712 — Drag-select marking stops when the cursor leaves the checkbox column.

### Root Cause

The `onMouseEnter` handler that triggers drag selection was only attached to the checkbox `<TableCell>` in `AlertRow.tsx`. When the cursor drifts outside the checkbox column while dragging down, the browser fires `onMouseLeave` on the checkbox cell but no `onMouseEnter` on adjacent cells, so the drag selection never fires for rows the cursor passes over outside the checkbox column.

### Fix

Added a row-level `onMouseEnter` handler (`handleRowMouseEnter`) to the `<TableRow>` element that checks `isDragging` before calling `onDragEnter`. This ensures drag-selection fires regardless of which column the cursor is over.

### Changes

- `apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx`
  - Added `handleRowMouseEnter` handler that fires `onDragEnter` during active drag
  - Added `onMouseEnter={handleRowMouseEnter}` to the `<TableRow>` element

### Testing

- Drag-select from checkbox down through rows: ✓ works across entire row width
- Single-click selection (no drag): ✓ unchanged
- Click on row content (not checkbox): ✓ unchanged

Closes #712

## Summary by CodeRabbit

* **New Features**
  * Enhanced drag selection in the alerts table, allowing users to select alerts by dragging across an entire row.
  * Drag interactions now respond when hovering over any part of a row, not only the checkbox area.

### DIFF
--- a/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
+++ b/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
@@ -80,9 +80,12 @@ export const AlertRow = ({
 	};
 
 	const handleRowMouseEnter = () => {
-		// During drag selection, entering any part of the row should trigger
-		// the drag selection handler, not just the checkbox column (#712).
-		if (isDragging && onDragEnter) {
+		// During drag selection, entering any part of the row should trigger the drag
+		// selection handler, not just the checkbox column (#712). No isDragging guard:
+		// isDragging only turns true after the first drag-enter, so gating on it would
+		// ignore the first row the cursor reaches after leaving the checkbox column.
+		// The hook itself no-ops unless the mouse went down on a checkbox.
+		if (onDragEnter) {
 			onDragEnter(alert);
 		}
 	};
--- a/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
+++ b/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
@@ -79,6 +79,14 @@ export const AlertRow = ({
 		}
 	};
 
+	const handleRowMouseEnter = () => {
+		// During drag selection, entering any part of the row should trigger
+		// the drag selection handler, not just the checkbox column (#712).
+		if (isDragging && onDragEnter) {
+			onDragEnter(alert);
+		}
+	};
+
 	return (
 		<TableRow
 			className={cn(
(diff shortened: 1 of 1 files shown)


idanlodzki:
Thanks @waterWang — great diagnosis, and the second contribution from you this week! You found exactly the right root cause: `onMouseEnter` living only on the checkbox cell, and a row-level handler is exactly the right fix.

I tested the branch hands-on with a headless-browser drag simulation and found one subtle timing gap in the guard, so I pushed a small follow-up commit (e9b9b31) on top of your work rather than bounce it back to you:

**The gap:** `handleRowMouseEnter` was guarded with `if (isDragging && ...)`, but `isDragging` only becomes `true` *after* the first successful drag-enter — and before your change, that first enter could only come from a checkbox cell. So the exact sequence from #712 (press a checkbox, drift out of the column immediately, then drag down) still failed: the row-level handler saw `isDragging === false` for every row and never fired.

Measured on your branch before/after the follow-up:

| Scenario | Before | After |
|---|---|---|
| Press checkbox → drift into row body → drag down 3 rows | 1 of 4 selected | **4 of 4** |
| Press checkbox → enter next checkbox → drift out → drag down | 5 of 5 | 5 of 5 |

**Why dropping the guard is safe:** `useDragSelection`'s `handleDragEnter` already returns early unless the mouse actually went down on a checkbox (`dragStateRef.current.isMouseDown`), so plain hovering over rows stays a no-op — the guard was duplicating a check the hook already makes, just with worse timing.

The fix keeps your handler and comment structure intact — it's a one-condition change. Thanks again for digging into this one; verified working in both the normal table and the split assigned/unassigned view.

PR #732, opened 2026-08-02
feat: persist split-by-owner and severity-colors toggles with the dashboard

### LINKED ISSUE #714: [Feature] Save "Split by owner" and "Severity colors" toggles with the dashboard
**Is your feature request related to a problem? Please describe.**
The alerts toolbar's "Split by owner" and "Severity colors" toggles reset on every visit: split-by-owner is plain component state in `Alerts.tsx` (`useState`), and severity colors persist only per-browser (`useSeverityColors` / localStorage). Neither is part of the dashboard, so a saved dashboard doesn't reproduce the view the user actually works with.

**Describe the solution you'd like**
Persist both toggles as dashboard fields, following the established pattern (#707: shared type + zod schema + `dashboards` table column with guarded migration + dirty-state tracking + load-on-select):
- `splitByAssignment: boolean`
- `severityColors: boolean`

**Additional context**
Consider precedence when both a dashboard value and the legacy localStorage severity-colors value exist — dashboard should win when one is loaded; the localStorage value can remain the fallback for unsaved drafts.

waterWang:
Store the 'Split by owner' and 'Severity colors' toolbar toggles as dashboard fields so a saved dashboard reproduces the exact view the user worked with. Follows the established pattern from #707.

Closes #714

## Summary by CodeRabbit

* **New Features**
  * Dashboard preferences for “Split by owner” and “Severity colors” are now saved and restored automatically.
  * Toolbar controls update the active dashboard’s saved settings.
  * Preferences remain available when dashboards are viewed in TV mode.
* **Bug Fixes**
  * Dashboard preference changes are now included when saving and detecting unsaved changes.
  * Existing dashboards continue to work with the new settings.

### DIFF
--- a/apps/client/src/components/Alerts/Alerts.tsx
+++ b/apps/client/src/components/Alerts/Alerts.tsx
@@ -46,7 +46,6 @@ import {
 	useAlertTagKeys,
 	useResolvedTabStatusFilterReset,
 	useColumnManagement,
-	useSeverityColors,
 } from './hooks';
 
 // Options for the alert-list picker (one dropdown instead of three toggle buttons);
@@ -86,8 +85,14 @@ const Alerts = () => {
 	const [filterPanelCollapsed, setFilterPanelCollapsed] = useState(false);
 	const [showDashboardSettings, setShowDashboardSettings] = useState(false);
 	const [pendingAction, setPendingAction] = useState<PendingAlertAction | null>(null);
-	const [splitByAssignment, setSplitByAssignment] = useState(false);
-	const { severityColors, toggleSeverityColors } = useSeverityColors();
+	// Read "Split by owner" toggle from dashboard state (persisted with the dashboard).
+	// Falls back to false when no dashboard is loaded.
+	const splitByAssignment = dashboardState.splitByAssignment;
+	const handleToggleSplitByAssignment = () => updateDashboardField('splitByAssignment', !splitByAssignment);
+	// Read "Severity colors" toggle from dashboard state, with localStorage as fallback
+	// for unsaved drafts (legacy behavior).
+	const severityColors = dashboardState.severityColors ?? (localStorage.getItem('opsimate-alerts-severity-colors') === 'true');
+	const handleToggleSeverityColors = () => updateDashboardField('severityColors', !severityColors);
 
 	const allAlerts = useMemo(() => [...alerts, ...resolvedAlerts], [alerts, resolvedAlerts]);
 	const tagKeys = useAlertTagKeys(allAlerts);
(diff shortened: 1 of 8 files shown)


idanlodzki:
Thanks @waterWang — you picked the right pattern to follow (#707's `column_order`: optional Zod field, guarded `ALTER TABLE`, `?? default` on load), and the client-side wiring through `DashboardContext` (state field, unsaved-changes detection, `userEditableFields`) is exactly how it should be structured. Three things need fixing before this can go in — the first one is a hard blocker:

### 1. 🔴 Blocker: better-sqlite3 cannot bind booleans — every dashboard save will crash

In `dashboardRepository.ts`, `createDashboard` and `updateDashboard` bind the values directly:

```ts
dashboard.splitByAssignment ?? null,
dashboard.severityColors ?? null,
```

better-sqlite3 only binds numbers, strings, bigints, buffers, and null — a JS boolean throws. I reproduced it against the repo's own dependency:

```
TypeError: SQLite3 can only bind numbers, strings, bigints, buffers, and null
```

And since `DashboardState` defaults both fields to `false`, the client sends a boolean on **every** save — so this isn't an edge case; creating or updating any dashboard 500s. Convert at the boundary:

```ts
dashboard.splitByAssignment == null ? null : dashboard.splitByAssignment ? 1 : 0,
dashboard.severityColors == null ? null : dashboard.severityColors ? 1 : 0,
```

### 2. Read path: SQLite returns `0`/`1`, not `false`/`true`

The mirror problem in `mapRowToDashboard`:

```ts
splitByAssignment: dashboardRow.split_by_assignment ?? undefined,
```

passes the raw INTEGER through, so the API returns `0`/`1` where the client types promise `boolean` (it happens to work today only because the client uses them in truthy positions). Convert on read and fix the `DashboardRow` model to be honest about what's in the DB:

```ts
// dashboardRepository.ts
splitByAssignment: dashboardRow.split_by_assignment == null ? undefined : dashboardRow.split_by_assignment === 1,

// models.ts
split_by_assignment?: number | null; // 0/1 boolean
severity_colors?: number | null; // 0/1 boolean
```

(Note `?? undefined` also has a subtle bug on its own: a stored `0` isn't nullish, so it comes through as `0`, not `false`.)

### 3. The localStorage fallback for severity colors is dead code, and the old preference is silently dropped

In `Alerts.tsx`:

```ts
const severityColors = dashboardState.severityColors ?? (localStorage.getItem('opsimate-alerts-severity-colors') === 'true');
```

`DashboardState.severityColors` is a non-nullable `boolean` defaulted to `false`, so it is never nullish and the `??` arm can never execute. Two consequences: users who had severity colors on via the old `useSeverityColors` hook lose their preference, and toggling without saving a dashboard no longer survives a reload (the old hook wrote to localStorage; nothing writes it now).

Suggested shape that keeps both behaviors: seed the default from localStorage and keep writing it on toggle, with the dashboard value overriding on load —

```ts
// DashboardContext.tsx defaultState
severityColors: localStorage.getItem('opsimate-alerts-severity-colors') === 'true',
```

```ts
// Alerts.tsx
const severityColors = dashboardState.severityColors;
const handleToggleSeverityColors = () => {
	localStorage.setItem('opsimate-alerts-severity-colors', String(!severityColors));
	updateDashboardField('severityColors', !severityColors);
};
```

Then drop the now-unused `useSeverityColors` hook (and its export in `hooks/index.ts`) since this replaces it.

---

Everything else checks out: the guarded `PRAGMA table_info` migration is correct, using `CreateDashboardSchema` covers both the create and update routes (the update controller parses with the same schema), the TV-mode save payload is updated, and the unsaved-changes detection wiring is right. With the three fixes above this closes #714 nicely.

PR #735, opened 2026-08-02
fix: stop the alert name column from ballooning on wide screens

idanlodzki:
## Problem

`alertName` and `summary` were the only two `w-auto` columns in the layout-fixed alerts table. Under CSS fixed table layout, all leftover width is split **equally** among auto columns — so on wide screens the Alert Name column ballooned with whitespace while Summary (the column with actually-long content) couldn't use the space.

Measured on main:

| Viewport | Alert Name | Summary |
|---|---|---|
| 1440px | 211px | 211px |
| 2560px | **771px** | 771px |

Alert names are typically 15–30 chars (~150–250px), so ~500px of the name column was dead space.

## Fix

- Pin `alertName` at `w-[240px]`; `summary` becomes the single flexible column and absorbs all surplus. Long names already truncate with a hover tooltip, and the expand-rows toggle already wraps them.
- New `getColumnWidthClass(column, orderedColumns)` helper: when Summary is hidden, `alertName` falls back to `w-auto` so the table always keeps exactly one flexible column. Both the header and body rows (separate tables in this architecture) size through the helper — diverging classes would shift them out of alignment.
- `COLUMN_MIN_WIDTHS.alertName` raised to 240 to match.

## Verification (live, headless Chromium)

| Config | Result |
|---|---|
| 2560px default | Alert Name 240px, Summary **1302px** (was 771/771) |
| 1440px default | Alert Name 240px, Summary 182px |
| Summary hidden | Alert Name becomes flexible: 422px @1440, 1542px @2560 |
| Both hidden | fixed columns stretch proportionally — no 0px collapse, no overlap |

Header/body cell alignment verified in every config. Client tests 47/47; tsc/lint/format at baseline.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alert table column sizing for more consistent alignment.
  * Increased the minimum width of alert names to improve readability.
  * Adjusted column behavior when the summary column is hidden.

### DIFF
--- a/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
+++ b/apps/client/src/components/Alerts/AlertsTable/AlertRow/AlertRow.tsx
@@ -1,9 +1,10 @@
+import { Fragment } from 'react';
 import { Checkbox } from '@/components/ui/checkbox';
 import { TableCell, TableRow } from '@/components/ui/table';
 import { cn } from '@/lib/utils';
 import { extractTagKeyFromColumnId, isTagKeyColumn } from '@/types';
 import { Alert } from '@OpsiMate/shared';
-import { ACTIONS_COLUMN, COLUMN_WIDTHS, getColumnWidthClass, SELECT_COLUMN_WIDTH } from '../AlertsTable.constants';
+import { ACTIONS_COLUMN, COLUMN_WIDTHS, SELECT_COLUMN_WIDTH } from '../AlertsTable.constants';
 import { CELL_PADDING } from './AlertRow.constants';
 import { getAlertSeverity, SEVERITY_ROW_CLASSES } from '../../utils/severity.utils';
 import { AlertActionsColumn } from './Columns/AlertActionsColumn';
@@ -151,7 +152,7 @@ export const AlertRow = ({
 								key={column}
 								alert={alert}
 								expanded={expandRows}
-								className={getColumnWidthClass('alertName', orderedColumns)}
+								className={COLUMN_WIDTHS.alertName}
 							/>
 						);
 					case 'severity':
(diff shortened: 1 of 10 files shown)


idanlodzki:
Reworked after testing with real dashboards: the alertName→flexible fallback just moved the ballooning problem (with Summary hidden, Alert Name absorbed the entire surplus while tag columns truncated at 110px next to it). Now:

- `alertName` is always fixed at 240px.
- When Summary is hidden, an **empty filler column** before the actions column absorbs the leftover width — the table reads left-packed with blank space on the right, like a normal data table, instead of inflating whichever data column happens to be flexible.
- Tag/default columns widened 110px → 150px so typical tag values (`production-east…`) fit; longer ones still truncate with a hover tooltip.

Verified at 1440/2000/2560px: default view unchanged (Summary absorbs all surplus), Summary-hidden + tag column now renders Name 240 / tag 150 / filler 592 / actions, header–body alignment holds in every config.

idanlodzki:
Second rework, after feedback on real dashboards: fixed widths + filler didn't use the space well — values truncated while blank space sat next to them.

Now the name and tag columns are **content-aware**: the longest visible value in each column is measured (canvas text metrics, bold weight since unread rows render bold) and the column gets exactly that width — clamped by a shrink floor (170px name / 110px tag) and a growth cap (460px / 480px). If the content columns collectively don't fit, they shrink proportionally to their slack (`distributeWidths`, 7 new unit tests); Summary (when visible) and the filler absorb only what content leaves over. Header and virtualized rows both receive the same computed pixel widths via inline styles, so the separate tables stay aligned by construction.

Measured live:

| Config | Result |
|---|---|
| 2000px, Summary hidden + Pod tag | Name 291px (longest name fits), Pod 406px (full pod name fits), **zero truncated cells** |
| 2000px default | Name 291px, Summary absorbs 691px |
| 1100px (tight) | columns shrink to floors, truncation + tooltips, header/body aligned |

Widths recompute automatically on pane/window resize (ResizeObserver) and when the alert list changes.

idanlodzki:
Follow-up fix for a phantom horizontal scrollbar that appeared under the header (header appeared scrollable while the body stayed put).

**Cause:** the body scrollport reserves a vertical-scrollbar gutter (~15px on classic-scrollbar systems), mirrored by the header via `scrollbar-gutter: stable` — so rows get ~15px less width than the outer scroller. The width hook measured the outer scroller, making the computed columns overflow the header table by exactly the gutter; the overflow leaked into the shadcn `Table` wrapper's own `overflow-auto` div, which rendered its own scrollbar right under the header. The body tables sit behind `overflow-x-hidden`, so they clipped silently — hence "only the header scrolls".

**Fix:** measure the scrollport's `clientWidth` (excludes the gutter), `floor()` instead of `round()` when shrinking columns so the summed widths can never exceed the budget, plus an 8px safety buffer.

Verified on the reproducing config (Started At + Last Update + Owner + 3 tag columns, no name/summary): header and body tables now both exactly match the container — no scrollable element left. All previous width scenarios re-measured unchanged.

PR #744, opened 2026-08-04
fix: kill the alerts header's own scrollbar and its clipped right edge

idanlodzki:
## Bug

With enough visible columns (see the "tester" dashboard) the table header grew **its own horizontal scrollbar** right under the column labels, could be scrolled independently of the body, and the header/actions area ended clipped at the right edge — even scrolled fully right, the last ~15px stayed hidden.

## Root causes (classic-scrollbar systems, e.g. macOS "always show scroll bars")

1. **Nested scroller:** the header used the ui/`Table` component, which wraps its `<table>` in an `overflow-auto` div → a second, independently scrollable region. The body rows had already dropped that wrapper; the header hadn't.
2. **Gutter not budgeted:** content-aware column widths distributed the scroller's *full* width, but the body scrollport reserves a scrollbar gutter (`scrollbar-gutter: stable`) — so the columns always overflowed by exactly one scrollbar width: permanent horizontal scrollbar + clipped actions column.

## Fix

- Header renders a raw `<table>` — it physically cannot scroll on its own; only the shared horizontal scroller moves header + body together.
- The width budget subtracts the measured system scrollbar width (`getScrollbarWidth()`, 0 on overlay-scrollbar systems), and the scroller's min-width floor adds it back so the column minimums fit *inside* the gutter.
- **Expand rows moves to the page toolbar**, between "Severity colors" and the time picker — a page-level preference persisted in localStorage (`useExpandRows`, mirroring `useSeverityColors`) that now drives every table on the page. The actions header column shrinks accordingly (80→44px, 116→80px with column settings).

## Verified live on the tester dashboard

- DOM probe: exactly **one** horizontally scrollable element remains (was two).
- Scrolled fully right: group-by/columns buttons and row action menus fully visible, header/body aligned, no clipping.
- Expand rows toggles from the toolbar, wraps long labels, persists across reloads.
- Typecheck, build, prettier, lint clean; client tests 55/55 pass.

## Summary by CodeRabbit

* **New Features**
  * Added an “Expand rows” control to the Alerts page, allowing wrapped alert content across all alert views.
  * Preserved row-expansion preferences between visits.

* **Improvements**
  * Improved Alerts table scrolling and sticky headers for easier navigation.
  * Adjusted action-column sizing for a more consistent layout.
  * Streamlined table controls by grouping related actions and column settings.

### DIFF
--- a/apps/client/src/components/Alerts/Alerts.tsx
+++ b/apps/client/src/components/Alerts/Alerts.tsx
@@ -22,7 +22,7 @@ import { useServices } from '@/hooks/queries/services';
 import { useToast } from '@/hooks/use-toast';
 import { cn } from '@/lib/utils';
 import { Alert } from '@OpsiMate/shared';
-import { Bell, CheckCircle2, ChevronDown, Columns2, LayoutList, Palette } from 'lucide-react';
+import { Bell, CheckCircle2, ChevronDown, Columns2, LayoutList, Palette, WrapText } from 'lucide-react';
 import { useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { AlertsFilterPanel } from '.';
@@ -45,6 +45,7 @@ import {
 	useAlertsRefresh,
 	useAlertTagKeys,
 	useColumnManagement,
+	useExpandRows,
 	useSeverityColors,
 } from './hooks';
 
@@ -87,6 +88,7 @@ const Alerts = () => {
 	const [pendingAction, setPendingAction] = useState<PendingAlertAction | null>(null);
 	const [splitByAssignment, setSplitByAssignment] = useState(false);
 	const { severityColors, toggleSeverityColors } = useSeverityColors();
+	const { expandRows, toggleExpandRows } = useExpandRows();
 
 	const allAlerts = useMemo(() => [...alerts, ...resolvedAlerts], [alerts, resolvedAlerts]);
 	const tagKeys = useAlertTagKeys(allAlerts);
@@ -397,6 +399,7 @@ const Alerts = () => {
 			onSearchTermChange={(term) => updateDashboardField('query', term)}
 			renderToolbar={false}
 			severityColors={severityColors}
+			expandRows={expandRows}
 		/>
 	);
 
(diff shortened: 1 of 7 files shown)


apps/client/src/components/Alerts/AlertsTable/AlertsTable.constants.ts
```diff
@@ -1,12 +1,12 @@
...
-// padding. When the column-settings button is also shown the header needs a third button,
-// so the table switches to the wider constant — with only 80px the buttons overflow the
-// fixed <th> and sit on top of the neighboring column's header text.
-export const ACTIONS_COLUMN_WIDTH = '80px';
-export const ACTIONS_COLUMN_WIDTH_WITH_SETTINGS = '116px';
+// Sized for the group-by header button: 28px + 16px cell padding. When the
+// column-settings button is also shown the header needs a second button, so the table
+// switches to the wider constant — with only 44px the buttons overflow the fixed <th>
+// and sit on top of the neighboring column's header text.
+export const ACTIONS_COLUMN_WIDTH = '44px';
```
idanlodzki, line 8: Minor: `COLUMN_MIN_WIDTHS[ACTIONS_COLUMN]` (line 76 below) still says `80` with a comment claiming it mirrors the inline-styled width, but the column is now 44px (80px only with column settings). The entry is currently dead — both consumers (`tableMinWidth` in AlertsTable.tsx and `claimed` in useContentColumnWidths) special-case ACTIONS_COLUMN via `actionsColumnWidth` — so nothing breaks today, but any future `COLUMN_MIN_WIDTHS['actions']` lookup would silently over-reserve 36px. Suggest deleting the entry (or deriving it from `ACTIONS_COLUMN_WIDTH`) while you're renumbering these constants.

apps/client/src/components/Alerts/Alerts.tsx
```diff
@@ -609,6 +612,17 @@ const Alerts = () => {
 									<span className="hidden lg:inline">Severity colors</span>
 								</Button>
 
+								<Button
+									variant={expandRows ? 'default' : 'outline'}
+									size="sm"
+									onClick={toggleExpandRows}
+									className="gap-1.5 shrink-0"
+									title="Expand rows to show full content"
```
idanlodzki, line 620: Small a11y regression vs the control this replaces: the old header button exposed `aria-pressed={expandRows}` and a state-aware label ("Expand rows" / "Collapse rows"); here the pressed state is conveyed only by the `variant` color, and the `title` stays "Expand rows to show full content" even when active. Suggest adding `aria-pressed={expandRows}` (the neighboring severity-colors/split toggles would benefit from the same, but this PR is the one removing an accessible toggle).

idanlodzki (commented):
Automated fresh-context review (no prior knowledge of this work). I verified the gutter math end-to-end (header table, body rows, useContentColumnWidths, tableMinWidth all agree after the subtraction), that the 44px actions column still fits the row menu button and the group-by header button, getScrollbarWidth caching/edge cases, the raw <table> swap vs ui/Table, and useExpandRows against the useSeverityColors pattern. No substantive bugs found; two minor notes below.

PR #784, opened 2026-08-07
[FIX]: Align tag length validation message

### LINKED ISSUE #765: Validation message says "less than 50 characters" but the rule allows exactly 50
## Problem
In `apps/client/src/components/ValidationFeedback.tsx` (lines ~142-144), the `tag-length` rule is labelled `Must be less than 50 characters` while the validator is `value.length <= 50`. A 50-character tag name passes validation even though the copy says it should not.

## Suggested fix
Change the label to "Must be 50 characters or less" (or tighten the validator to `< 50` — pick whichever matches the intended limit).

jiahao6635:
## Issue Reference

Closes #765

---

## What Was Changed

- Updated the tag-length validation copy to say "Must be 50 characters or less".
- Centralized the inclusive 50-character limit and its label.
- Added regression coverage for the label and the 50/51-character boundary.

---

## Why Was It Changed

The validator accepts exactly 50 characters, while the previous message said values must be less than 50 characters. The updated copy now matches the actual inclusive limit, and the shared constant helps keep the validator and message aligned.

---

## Screenshots

Not applicable. The `tagName` validation rule is currently not referenced by a rendered client view, so there is no existing UI flow that can display this message. The regression test verifies the production rule directly.

---

Validation completed:

- `pnpm build` (4/4 workspace packages passed)
- Client test suite (10 files, 62 tests passed)
- Client lint (0 errors; pre-existing warnings only)
- Prettier check and `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized tag-name validation to enforce a maximum of 50 characters.
  - Updated the validation message to clearly state the character limit.

- **Tests**
  - Added coverage confirming that 50-character names are accepted and 51-character names are rejected.

### DIFF
--- a/apps/client/src/components/ValidationFeedback.constants.ts
+++ b/apps/client/src/components/ValidationFeedback.constants.ts
@@ -1,0 +1,2 @@
+export const TAG_NAME_MAX_LENGTH = 50;
+export const TAG_NAME_MAX_LENGTH_LABEL = `Must be ${TAG_NAME_MAX_LENGTH} characters or less`;
--- a/apps/client/src/components/ValidationFeedback.test.ts
+++ b/apps/client/src/components/ValidationFeedback.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from 'vitest';
+import { validationRules } from './ValidationFeedback';
+
+const EXPECTED_TAG_NAME_MAX_LENGTH = 50;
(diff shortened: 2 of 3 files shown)


idanlodzki (approved):
Reviewed: label and validator now agree via a shared constant, plus a regression test pinning the inclusive 50-char limit. LGTM — thanks for the contribution!

PR #822, opened 2026-08-10
fix: enrich resolved alerts the same way active ones are enriched

idanlodzki:
## Why they weren't

`AlertBL.getAllAlerts()` enriches before returning:

```ts
let alerts = await this.alertRepo.getAllAlerts();
if (this.enrichmentBL) alerts = await this.enrichmentBL.applyEnrichments(alerts);
```

`getAllResolvedAlerts()` only attached comments and firing times — it never called `applyEnrichments`. Nothing was disabling enrichment for resolved alerts; that code path simply never ran it.

Measured on one alert, before and after resolving it, with a rule that adds a link and rewrites a tag:

| | `links` | `tags` |
|---|---|---|
| active | `[{label: 'dashobard', …}]` | `{service: '👍kafka'}` |
| resolved (before this fix) | `null` | `{service: 'kafka'}` |
| resolved (after) | `[{label: 'dashobard', …}]` | `{service: '👍kafka'}` |

## Actions

Not a second bug — the same gap. Action filters match on `alert.tags` (`actionMatchesAlert` in `AlertActions.tsx`), with no status check anywhere, so:

- an action keyed to a **plain** tag or name always worked on resolved alerts
- an action keyed to an **enrichment-added** tag stopped matching the moment the alert resolved, because the alert arrived unenriched
- worse, `runOnAlert` builds its context from the alert the client passes in, so an action run from the Resolved view would have templated its payload from raw values

Enriching the read path fixes all three.

## What I deliberately did NOT change

**Mute policies stay active-only.** `markMuted` is still not applied to resolved alerts: muting suppresses a *firing* alert, so "Muted" carries no meaning once it's resolved, and the Resolved view suspends the status filter anyway. A test pins this so the asymmetry is intentional and documented rather than looking like the same oversight.

## Verification

- 140 server tests (2 new): a resolved alert carries the same enrichment an active one gets (templated tag + link + `appliedEnrichments`), and a mute policy does **not** mark a resolved alert muted
- Mutation-tested: removing the `applyEnrichments` call makes the new test fail, so it genuinely pins the fix
- Live against the real API — the same alert resolved before/after a server restart produced the table above, with `appliedEnrichments: [{id: 13, …}, {id: 14, …}]`
- Live in the UI: the resolved alert's details panel now shows the enrichment label `service: 👍kafka`, the enrichment link, and its matching Actions

## Summary by CodeRabbit

* **New Features**
  * Dashboard alert settings now save and restore assignment splitting and severity-color preferences.
  * Existing severity-color preferences are preserved when dashboard settings are not yet configured.
  * Dashboard APIs support validating and persisting the new alert settings.

* **Bug Fixes**
  * Resolved alerts now retain enrichment details while continuing to exclude mute policies.

### DIFF
--- a/apps/server/src/bl/alerts/alert.bl.ts
+++ b/apps/server/src/bl/alerts/alert.bl.ts
@@ -270,9 +270,19 @@ export class AlertBL {
 	async getAllResolvedAlerts(): Promise<Alert[]> {
 		try {
 			logger.info('Fetching all resolved alerts');
-			return await this.attachLastComments(
-				await this.attachFiringTimes(await this.resolvedAlertRepo.getAllResolvedAlerts())
-			);
+			let alerts = await this.resolvedAlertRepo.getAllResolvedAlerts();
+			// Resolved alerts are enriched exactly like active ones. The added tags, links,
+			// summary and severity are how an alert is DESCRIBED, and that description
+			// shouldn't change the instant it resolves: the Resolved view would show raw
+			// values, and any action whose filter matches an enrichment-added tag would
+			// stop offering itself (and would template its payload from unenriched values).
+			//
+			// Mute policies are deliberately not applied here: muting suppresses a firing
+			// alert, so "Muted" carries no meaning once the alert is resolved.
+			if (this.enrichmentBL) {
+				alerts = await this.enrichmentBL.applyEnrichments(alerts);
+			}
+			return await this.attachLastComments(await this.attachFiringTimes(alerts));
 		} catch (error) {
 			logger.error('Error fetching resolved alerts', error);
 			throw error;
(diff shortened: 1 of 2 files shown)


idanlodzki:
@coderabbitai review

idanlodzki:
Rebased onto main — my mistake: I branched this off the #821 branch instead of main, so the PR was carrying #821's commits and the diff showed 14 files. Now that #821 is merged, this PR is just the two server files it describes (`alert.bl.ts` + its tests), one commit.

That also explains the nitpick above: `dashboardToggles.test.tsx` is #821's file, not part of this change, so it's out of scope here — noting it so it isn't lost.

@coderabbitai review

PR #830, opened 2026-08-12
feat(alerts): bulk actions by filter — act on every matching alert in one request (Phase 2b)

idanlodzki:
## What

Phase 2b of the alerts scalability epic (after #827 snapshot cache, #828 server-side querying, #829 group summaries).

Bulk actions on alerts used to fan out **one HTTP request per alert** and could only touch **loaded rows** — the selection bar even warned "actions apply to the loaded alerts only; scroll to load more first". On a 10K instance that meant silencing a filtered set was either 500 requests (one page) or impossible without scrolling everything in.

### Server
- **`POST /api/v1/alerts/bulk`** — actions `silence | unsilence | resolve | assignOwner | comment`, scoped by **exactly one** of:
  - `ids` — the loaded selection (one request instead of N; capped at 10k per request)
  - `query` — the same `filters`/`from`/`to`/`search` shapes the list endpoints speak, resolved server-side against the full dataset with the shared engine, so "act on all N matching" agrees exactly with what the view shows
- The BL **loops the existing single-alert methods** — history events, ownership takeover on silence/resolve, and comment side-effects stay identical to acting one-by-one. Zero new mutation semantics.
- Honest counts: response is `{matched, succeeded, failed}`; unknown ids count as **failed** (the single methods return null rather than throwing), and `resolveAlert` now returns whether it actually resolved.
- `express.json` limit raised to 2mb (default 100kb rejects a few thousand ids).

### Client
- **Gmail pattern** in the selection bar: select the whole loaded page while the server counts more matching → "**Select all 2,500 matching**". Arming it banners the scope ("actions apply to every match, loaded or not") and widens every bulk action to the query.
- The N comes from a limit-1 list request (`useAlertMatchCount`) — same engine, true total, fetched only while a selection is open.
- Confirmation dialogs say the real number and call out the widened scope explicitly.
- **Permanent delete stays loaded-selection only** on purpose — an irreversible action never applies to alerts the user hasn't had a chance to see; the button hides while "all matching" is armed.
- Regular selections also go through the bulk endpoint with ids — silencing a 500-row selection is now **1 request, not 500**.
- Armed scope drops on any selection/filter/search/window/tab change — the N the user confirmed must stay the N that runs.
- Playground mock speaks the same contract through shared single-op helpers (no drift), and its list routes now honor the query contract (paged responses with true totals).

## Verified live on the 10K instance
- Selected 500 loaded of 2,500 matching "Memory" → "Select all 2,500 matching" → armed banner → Silence all → dialog "Silence 2500 alerts?" with scope note → **one** `POST /alerts/bulk` → toast "Silenced 2500 alerts", sidebar Silenced 2500, every row silenced.
- Restore via bulk unsilence-by-query: 2,500 alerts in **0.4s**.
- ids path: 2-row selection → Comment → one request → "Commented on 2 alerts".

## Tests
- 12 new server tests (`alertsBulkApi.test.ts`): id/query scoping, search parity with the list endpoint, honest failed counts for unknown ids, empty-query = whole list, all validation 400s.
- 201 server + 145 client tests green, lint 0 errors.

## Trade-offs
- A bulk action is a loop of single-alert mutations server-side (SQLite ops are sync/fast; 2,500 silences incl. history events + ownership ≈ instant, restore measured at 0.4s). Write-time batching belongs to Phase 2c if ever needed.
- `assignOwner` and `unsilence` run without a confirm dialog (matching their existing loaded-selection behavior) even when "all matching" is armed — both are reversible.
- The match count uses the coarse (minute-rounded) time window the queries already use; on rolling presets the acted-on set is consistent with the displayed N by construction.

## Summary by CodeRabbit

* **New Features**
  * Added bulk alert actions for silencing, unsilencing, resolving, assigning, and commenting.
  * Select all matching alerts across filtered results, with match counts and scope confirmation.
  * Display success and failure counts for partially completed bulk actions.
* **Bug Fixes**
  * Prevented permanent deletion outside the currently loaded selection.
  * Reset selections when switching views or changing filters.
  * Improved validation for bulk action requests and targeting scopes.

### DIFF
--- a/apps/client/src/components/Alerts/Alerts.tsx
+++ b/apps/client/src/components/Alerts/Alerts.tsx
@@ -13,7 +13,9 @@ import { deserializeTimeRange, readLegacySeverityColors, serializeTimeRange } fr
 import {
 	useAlertFacets,
 	useAlertGroupSummaries,
+	useAlertMatchCount,
 	useAlerts,
+	useBulkAlertAction,
 	useResolvedAlerts,
 	useDeleteResolvedAlert,
 	useMarkAlertRead,
@@ -30,7 +32,7 @@ import { AlertFacetsResponse } from '@/lib/api';
 import { cn } from '@/lib/utils';
 import { Alert } from '@OpsiMate/shared';
 import { Bell, BellOff, CheckCircle2, ChevronDown, Columns2, LayoutList, Palette, WrapText } from 'lucide-react';
-import { useMemo, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { AlertsFilterPanel } from '.';
 import { AlertDetailsPanel } from './AlertDetails';
 import { AlertsSelectionBar } from './AlertsSelectionBar';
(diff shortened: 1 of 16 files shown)


idanlodzki:
All review rounds addressed:

**CodeRabbit round 1** (commit b994c3a + 1b23ef9): partial counts on chunk failure ✅, non-numeric limit 400 in the mock ✅, prettier ✅, bulk contract declared once in `@OpsiMate/shared` (no anonymous inline types) ✅, `ownerId` validated numeric ✅, mock validation parity with the schema ✅.

**Deferred by design — snapshot invalidation once per batch:** the bulk loop reuses the single-alert methods verbatim, which each invalidate internally; suppressing that requires plumbing a flag through five mutation paths. Invalidation itself is a cheap generation bump — the cost only materializes when a poller triggers a recompute mid-batch. Measured on a 10K instance: 2,500 mutations complete in ~0.4s, i.e. within a single poll interval, so at the scale this PR targets the double-compute window is negligible. Write-time batching is Phase 2c's remit.

**Opus adversarial review** (commit b994c3a): action-time window resolution (rolling presets no longer freeze at mount), membership-based select-all gate, resolved/history/comments cache invalidation, no orphan comments on unknown ids (+test), 400-not-401 for authorless comments, scoped 2mb body limit, ownership-takeover disclosure in the silence dialog, real users into the playground engine.

PR #835, opened 2026-08-14
[FIX]: Remove dead provider client remnants

### LINKED ISSUE #776: Remove empty leftover files useViewedAlerts.ts and CustomAlertsIcon.tsx
## Problem
Two files are 0 bytes and imported nowhere (leftovers from the custom-webhooks integration commit `e836cce`):
- `apps/client/src/hooks/useViewedAlerts.ts`
- `apps/client/src/components/icons/CustomAlertsIcon.tsx`

## Suggested fix
Delete both files.

floze-the-genius:
## Summary

- remove the unused provider and service client wrappers left after the provider UI removal
- keep the live tag endpoints under a dedicated `tagApi` and update every consumer
- remove the two dead permission helpers and the two zero-byte leftover files

## Why Was It Changed

The provider/services UI was removed in #815, but its API wrappers and permission helpers remained. Removing the whole `providerApi` verbatim would break the tag UI because the same object still owned the live tag methods, so this change separates those methods before deleting the dead provider/service surface.

Closes #776.

## Validation

- `pnpm build` — 4/4 workspace packages built successfully
- `pnpm --filter @OpsiMate/client lint` — exit 0 (existing warnings only)
- Prettier check on all modified non-empty files — passed
- `pnpm --filter @OpsiMate/client test:run` — 121 passed / 24 failed; an untouched `upstream/main` worktree produced the identical 121/24 result and the same `localStorage` environment failures
- `pnpm --filter @OpsiMate/client exec tsc -b --pretty false` — current and untouched `upstream/main` both report the same 34 pre-existing errors, with no errors in changed files

## AI Assistance

This contribution was implemented with OpenAI Codex. The current `main` references, merged provider-removal change, diff, and validation results were inspected directly; the AI-assisted scope includes code changes, validation, commit message, and this PR description.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag management reliability across creation, deletion, selection, and service-tag assignment.
  * Tag lists and service tags now load consistently through the dedicated tag service.
* **Refactor**
  * Consolidated tag-related operations under a dedicated API, preserving existing validation, loading states, error handling, and callbacks.
  * Removed obsolete provider and service permission checks.


idanlodzki (approved):
Reviewed by checking out the branch and running the full gate locally, plus tracing every removed symbol. **Approving** — the core judgment call here was the non-obvious one and you got it right.

Deleting `providerApi` verbatim would have broken the tag UI, because that object still owned the live tag methods. Splitting them into a dedicated `tagApi` *first*, then removing the dead provider/service surface, is exactly the right sequencing, and all four consumers were updated.

**What I verified independently:**

- **No dangling references** anywhere in the repo to `providerApi`, `serviceApi`, `canManageProviders`, `canViewServices`, `useViewedAlerts`, or `CustomAlertsIcon`.
- **Build and lint pass**; client `tsc` reports 34 errors — identical to `main`, none in the files you touched. Your pre-existing-error analysis was accurate.
- The tag endpoints you kept genuinely exist server-side in `apps/server/src/api/v1/tags/router.ts`, and `getAllTags` / `createTag` / `deleteTag` have live UI consumers.

---

**Two notes, neither blocking:**

**1. Your test numbers look like a local environment problem.** You reported `121 passed / 24 failed` and concluded (correctly) that it wasn't caused by your change since untouched `main` did the same. But running your branch here gives **145 passed / 0 failed**, matching both `main` and CI. Those `localStorage` failures are specific to your machine — worth fixing locally so future PRs get a clean signal, and worth correcting in the PR body so it doesn't read as if the suite is broken.

**2. The cleanup is incomplete, in the same category it set out to fix.** `tagApi` keeps three methods that call `/services/:id/tags`:

- `addTagToService`
- `removeTagFromService`
- `getServiceTags`

There is no services router mounted on the server at all (`v1.ts` mounts alerts, tags, dashboards, actions, … but nothing for services), so those calls would 404. Their only consumers are `TagSelector.tsx`, which nothing imports, and `useServiceTags`, which is only re-exported from the barrel — both orphaned by the same provider/services removal.

This is **not a regression** — that code was equally dead before your PR — so it doesn't block merging. Happy either way: extend this PR to remove those three methods plus `TagSelector.tsx` and `useServiceTags`, or leave it and I'll track it as a follow-up.

Please **update the branch from `main`** (it's behind by several merges) and this is good to go. Thanks for picking this up so quickly after the issue was scoped — nice, careful work.

PR #841, opened 2026-08-16
[FEAT]: unit tests for alertTags utils

### LINKED ISSUE #778: Add unit tests for alert tag helpers (alertTags.utils)
## What
`apps/client/src/components/Alerts/utils/alertTags.utils.ts` has no tests. Exports: `HIDDEN_TAG_KEYS`, `getAlertTagsArray`, `getAlertTagsString`, `getAlertPrimaryTag`, `hasAlertTags`, `alertMatchesTagFilter`, `getAlertTagEntries`.

## Behaviors to cover
- The `severity` key is always filtered out of tag lists
- Empty-string tag values are dropped
- A missing/non-object `tags` field returns `[]` instead of throwing
- `alertMatchesTagFilter` returns `true` for an empty filter list (no filter) and `false` when no tag matches
- `getAlertTagsString` joins with `", "` and returns `""` for no tags

## Setup
Vitest is already configured. Copy the pattern from `apps/client/src/test/distributeWidths.test.ts`. Run with `pnpm --filter @OpsiMate/client test`.

gshiva53:
## Issue Reference
Closes #778

---

## What Was Changed
Unit tests for alertTags utils that cover the following cases: 

#### `getAlertTagArray` and `getAlertTagEntries`
- The `severity` key is always filtered out of tag lists
- The `fix` key is always filtered out of tag lists
- Empty-string tag values are dropped
- A missing/non-object `tags` field returns `[]` instead of throwing

#### `alertMatchesTagFilter` 
- returns `true` for an empty filter list (no filter)
- returns `false` when no tag matches

#### `getAlertTagsString`
- joins with `", "`
- returns `""` for no tags

---

## Screenshots
Screenshot of the tests passing. 
image

---

## Additional Context 
This test file can be run using the following command from the OpsiMate directory.
```
cd apps/client
pnpm test src/test/alertTags.utils.test.ts
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage for alert tag filtering, matching, and formatting.
  * Added validation for empty, missing, reserved, and invalid tag values.


idanlodzki (changes requested):
Thanks @gshiva53 — the `getAlertTagsArray` and `getAlertTagsString` coverage here is good, and I confirmed it works: when I removed the filter from `getAlertTagsArray`, three of your tests failed immediately. That half is doing its job.

The `getAlertTagEntries` half, unfortunately, doesn't assert anything. Requesting changes for that one issue.

### The problem

`getAlertTagEntries` returns `Array<{ key: string; value: string }>` — **objects**, not arrays or strings. So this line:

```ts
getAlertTagEntries(testAlert).forEach((tagKV) => expect(tagKV).not.toContain('severity'));
```

passes even when `tagKV` **is** `{ key: 'severity', value: 'critical' }`. `toContain` on a plain object doesn't check what it looks like it checks here.

Two ways I confirmed it:

1. This test passes on its own:
```ts
expect({ key: 'severity', value: 'critical' }).not.toContain('severity'); // passes
```
2. I deleted the entire filter from `getAlertTagEntries` in `packages/shared` — no `severity`/`fix` filtering, no dropping of empty values — and **all 9 tests still passed**.

There's also a vacuity trap in the same lines: `forEach` over an empty array runs zero assertions, so if the function returned `[]` the test would pass too.

### The fix

Assert on the shape the function actually returns — map to the field you care about first:

```ts
it('filters out severity and fix keys from tag entries', () => {
	const keys = getAlertTagEntries(testAlert).map((entry) => entry.key);
	expect(keys).toEqual(['priority']); // exact, and fails if entries come back empty
});

it('drops empty-string tag values', () => {
	const values = getAlertTagEntries(testAlertWithEmptyTags).map((entry) => entry.value);
	expect(values).not.toContain('');
	expect(values).toHaveLength(1);
});
```

Asserting the full expected array (`toEqual`) rather than `not.toContain` is worth preferring throughout — it can't pass vacuously, and it fails loudly if the shape changes.

### Two smaller things while you're in there

- **Missing the primary case for `alertMatchesTagFilter`**: you cover the empty-filter and no-match cases, but not the one that matters most — returns `true` when a tag *does* match (e.g. `alertMatchesTagFilter(testAlert, ['P1'])`).
- `const emptyFilterValues = [];` is untyped; `const emptyFilterValues: string[] = []` states the intent.

Optional: the repeated `as unknown as Alert` casts would read better as a small `mkAlert(overrides)` factory — see `apps/server/tests/alertQueryEngine.test.ts` for that pattern.

Ping me when it's updated and I'll re-review. Also FYI: the `Run CI` workflow doesn't trigger automatically on fork PRs, so the suite you see passing here is only `CodeRabbit`/`label` — I ran the full client suite locally against your branch (155 tests, green).

gshiva53:
Thanks @idanlodzki for your review. I have made the following changes. 
- Updated the `getAlertTagArray` filter out test to use `toEqual` instead of `toContain`
- `mkAlert(overrides)` factory function instead of multiple `as unknown as Alert` type assertion
- primary case for `alertMatchesTagFilter`
- `emptyFilterValues` is now typed.

idanlodzki (approved):
This is exactly the fix — thanks @gshiva53. Approving and merging.

I re-ran the same mutation checks that caught the original problem, and the rewritten tests now catch every one of them:

| I broke... | Result |
|---|---|
| `getAlertTagEntries` — removed the hidden-key + empty-value filter | ✅ 2 tests fail (this is the one that passed silently before) |
| `alertMatchesTagFilter` — made it always return `true` | ✅ 1 test fails |
| `getAlertTagsString` — changed the separator to `" \| "` | ✅ 1 test fails |

Mapping entries to `.key` and asserting the exact array with `toEqual` is the right call — it can't pass vacuously the way `not.toContain` on an object did. The `mkAlert` factory and the positive `alertMatchesTagFilter` case are both nice additions. Full client suite green (160 tests), lint and prettier clean.

Two cosmetic things for next time, both non-blocking (not worth another round trip):

1. In `drops empty string tag values`, the variable is named `values` but maps `tag.key`, so `expect(values).not.toContain('')` is really asserting "no empty *key*". The test still catches the regression — `toHaveLength(1)` does that work, which I confirmed — but either rename it to `keys`, or map `tag.value` to make the assertion match the test's name.
2. Small typo: `fitlerValues` → `filterValues`.

Both issues (#778 and #779) are now closed. If you'd like more, the `good first issue` label is the place to look — and thanks for taking the review round seriously, the second pass was a genuine improvement.

PR #859, opened 2026-08-17
[FIX]: react router link in place of a html tag

### LINKED ISSUE #773: 404 page uses a raw <a href> instead of react-router Link
## Problem
The "Return to Home" link on `apps/client/src/pages/NotFound.tsx` (line ~19) is a plain `<a href="/">`, which triggers a full browser reload and discards React Query cache / app state. Every other navigation in the app uses react-router.

## Suggested fix
Replace with `<Link to="/">` — the file already imports from `react-router-dom`.

gshiva53:
## Issue Reference
Closes #773 

---

## What Was Changed
The `a` tag was changed to the `Link` from react router.

---

## Why Was It Changed
It was changed because of the reasoning provided in #773. I am not sure of how to test this functionality. 

---

## Screenshots
The `NotFound` page still renders the same way. 
Screenshot 2026-08-17 at 9 26 36 pm

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation from the Not Found page by using in-app routing to return home without a full page reload.
  * Added coverage for Not Found page rendering and home navigation.

### DIFF
--- a/apps/client/src/pages/NotFound.tsx
+++ b/apps/client/src/pages/NotFound.tsx
@@ -1,6 +1,6 @@
 import { Logger } from '@OpsiMate/shared';
 import { useEffect } from 'react';
-import { useLocation } from 'react-router-dom';
+import { useLocation, Link } from 'react-router-dom';
 
 const logger = new Logger('NotFound');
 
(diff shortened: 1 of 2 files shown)


idanlodzki (approved):
Approved and merging — correct fix, and thanks for catching the follow-up on the commented-out anchor yourself before I got here.

**On "I am not sure of how to test this functionality"** — good instinct to flag it, and it's a genuinely tricky one: react-router's `Link` renders a plain `<a href="/">`, so the rendered DOM is *identical* to what you replaced. Nothing about the markup can distinguish them.

The lever is that **jsdom doesn't implement real navigation**: clicking a bare `<a href="/">` can't change the router's location, but a `Link` can. So rendering the page inside a `MemoryRouter` and asserting the home route takes over separates the two with no mocking at all. I've pushed that test to your branch as `src/test/NotFound.test.tsx` — and I verified it the important way round: it **fails** against the old anchor version and passes against your `Link`. A test that can't fail on the old code wouldn't have been worth adding.

That pattern generalizes: when a change is about *behaviour that shares markup with the thing it replaces*, look for an observable the framework controls (here, router location) rather than something in the DOM.

For the record I also ran the full gate locally on your branch: 4/4 builds, 174 client tests, lint and prettier clean.

PR #860, opened 2026-08-17
[FIX]: add try-catch for error handling in testIntegration

### LINKED ISSUE #771: testIntegration and custom-actions handlers lack try/catch — validation errors become 500s
## Problem
Unlike every sibling handler, `POST /integrations/test-connection` (`apps/server/src/api/v1/integrations/controller.ts` ~34-56) has no error handling: a `CreateIntegrationSchema.parse` failure or connector exception falls through to Express's default error handler instead of returning `400 { success: false, error: "Validation error" }`.

The same gap exists in `apps/server/src/api/v1/custom-actions/controller.ts` (~56-59, ~61-66, ~87-91).

## Suggested fix
Wrap the handler bodies in the same try/catch + `isZodError` pattern used by `createIntegration` in the same file.

NBlaze860:
## Issue Reference

Closes #771 

---

## What Was Changed

Added try catch block to POST /integrations/test-connection ( testIntegration function ).

---

## Why Was It Changed

For better error handling.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved integration connection testing error handling.
  - Invalid request data now returns a clear HTTP 400 response with validation details.
  - Unexpected errors now return a consistent HTTP 500 response.
  - Existing successful and failed connection test responses remain unchanged.

### DIFF
--- a/apps/server/src/api/v1/integrations/controller.ts
+++ b/apps/server/src/api/v1/integrations/controller.ts
@@ -32,26 +32,35 @@ export class IntegrationController {
 	};
 
 	testIntegration = async (req: Request, res: Response) => {
-		const integrationToCreate = CreateIntegrationSchema.parse(req.body);
+		try {
+			const integrationToCreate = CreateIntegrationSchema.parse(req.body);
 
-		const integrationConnector = integrationConnectorFactory(integrationToCreate.type);
+			const integrationConnector = integrationConnectorFactory(integrationToCreate.type);
 
-		const integration: Integration = {
-			...integrationToCreate,
-			createdAt: '',
-			id: 0,
-		};
+			const integration: Integration = {
+				...integrationToCreate,
+				createdAt: '',
+				id: 0,
+			};
 
-		const testResult = await integrationConnector.testConnection(integration);
+			const testResult = await integrationConnector.testConnection(integration);
 
-		if (testResult.success) {
-			return res.status(200).json({ success: true, data: { isValidConnection: true } });
-		} else {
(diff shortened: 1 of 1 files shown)


idanlodzki (approved):
Approved and merging — thanks @NBlaze860, and welcome!

This is the right fix in the right shape: it mirrors the try/catch + `isZodError` pattern `createIntegration` uses a few lines up, so the handler now behaves like every one of its siblings instead of being the odd one out.

I verified it end to end rather than just reading it, on your branch and against `main`:

| Request | main (before) | your branch |
|---|---|---|
| Invalid body | **500 + an HTML error page** from Express's default handler | **400** `{"success":false,"error":"Validation error","details":[…]}` |
| Valid body, unreachable host | 200 `{isValidConnection:false}` | 200 `{isValidConnection:false}` — unchanged |

That second row is the part I most wanted to confirm: a connector that legitimately can't connect is a *result*, not a server error, and your catch correctly leaves that 200 contract alone rather than swallowing it into the 500 branch.

I also checked the wider claim behind "Closes #771". The issue mentions a second file (`custom-actions/controller.ts`) that no longer exists — the equivalent handlers now live in `custom-fields/controller.ts` and already have this pattern. I scanned every handler in `apps/server/src/api/v1/*/controller.ts` by brace-matching each body, and `testIntegration` was the **only** one left without a try/catch. So this genuinely closes the issue rather than half of it.

Full gate on your branch: 4/4 builds, 223 server tests, lint and prettier clean.

PR #888, opened 2026-08-23
feat(insights): hourly buckets for short windows + Today preset

idanlodzki:
## What

Two Insights improvements to the time axis.

### 1. Hourly buckets for short windows
The volume and trend charts bucketed by **day** regardless of window, so a 24h view was a single point. Now windows ≤ 48h bucket the series by **hour**; longer windows stay daily. The server picks the granularity and returns it on `range.granularity`, and the client formats the x-axis accordingly (`HH:00`, with the short date shown at midnight; `MMM D` for day buckets).

### 2. Dense axis
Bounded windows are now **zero-filled** across every bucket in `[from, to]`, so a 24h view is a continuous 24-point curve and quiet hours/days appear as gaps in the line instead of being dropped. (All-time stays sparse — it's unbounded.)

### 3. "Today" preset
Start of the local day → now (day-so-far), matching the alerts filters' *Today*. It's a short window, so it renders hourly automatically.

All of this applies to the overview volume chart, the MTTR/MTTA reliability trends, and the tag volume + per-value trend charts; chart hints adapt ("per hour" / "per day").

## Where

The bucketing lives in the shared `computeAlertAnalytics` (used by both the server endpoint and the playground), so the change is in `@OpsiMate/shared`.

## Tests & verification

- 30 analytics tests (2 new: hourly bucketing + dense fill; day-granularity retention). Full client suite 242, server analytics 30, all green. Client typecheck gate + lint + build clean.
- Verified **live** in the playground: 24h and Today render as proper hourly curves (evening peak visible), 7d+ stay daily with clean "Aug 17…Aug 24" labels.

## Summary by CodeRabbit

* **New Features**
  * Added hourly analytics for time ranges up to 48 hours, with daily views for longer periods.
  * Added a “Today” preset covering the current local day.
  * Updated volume, reliability, and tag trend charts with granularity-aware labels and intervals.
  * Ensured trend charts display complete time periods, including periods without measurements.

* **Bug Fixes**
  * Improved timezone-aware bucket generation, including daylight-saving transitions.
  * Aligned chart data across views for more consistent time-range comparisons.


idanlodzki:
## Review summary before merge

**CodeRabbit (round 1)** raised three findings, all fixed in `9f56794`:
- chart hints hardcoded "per day" on hourly views — now interpolate the active granularity across all five hints
- `nowBucket` first ticked at mount+5min, so Today could lag the midnight rollover — first tick now aligns to the next bucket boundary
- tag charts seeded only from buckets that had data, leaving a 24h tag view sparse — now seeded from the overview's dense bucket set

**Adversarial pass (round 2)** found one real bug CodeRabbit missed, fixed in `4ca28c8c`:

The dense-axis seed stepped a fixed 24h per day bucket. A local day is only 23 hours on the spring-forward date, so `t += 24h` from a late-evening `from` lands two calendar days later and that day is never seeded. A day carrying episodes still gets seeded by the episode loop, so what actually went missing was a **quiet** DST day — precisely the gap the dense axis exists to fill. Reproduced in Europe/Berlin, America/New_York, America/Santiago, Asia/Jerusalem and Australia/Sydney (7 of 8760 sampled 7-day windows per zone). Now steps hourly and lets repeats collapse in the Map; measured at ~14ms of cached Intl formatting over a two-year window. Covered by a regression test that fails without the fix.

Same commit swapped the last anonymous inline object type in the analytics test for the shared `DurationDayPoint`.

**Live verification** — 10K-alert scratch instance (12,374 episodes / 22,000 history rows / 17,350 events, Feb→now):

| preset | granularity | buckets | episodes |
|---|---|---|---|
| Today | hour | 23 | 620 |
| 24h | hour | 25 | 654 |
| 7d | day | 8 | 1,899 |
| 30d | day | 31 | 6,571 |
| 90d | day | 91 | 9,700 |
| All time | day | 190 | 12,374 |

The DST window renders `2026-03-26 … 2026-04-02` with `2026-03-29` present and no gaps. Narrowing to a single alert name over 24h gives 25 buckets with 5 non-zero — a full-width axis where the old behaviour drew five isolated points. Tag insights on 24h: 25 dense hourly buckets, all value-trends on one shared x-axis, zero-filled where a value had no data. Response cost at 10K: 0.09s (24h) / 0.16s (90d) / 0.19s (all time).

**Deferred, with rationale:**
- The three charts got a `tickFormatter` but their `<Tooltip>`s have no `labelFormatter`, so the axis reads `13:00` while the hover still shows `2026-08-30 13:00`. Cosmetic, and the full key is arguably the more useful thing on hover.
- `previousFrom = from - (to - from)` makes Today's KPI baseline a sliver of yesterday early in the morning. Equal-length preceding windows are the existing semantics for every preset; changing them for one preset is a separate discussion.

PR #890, opened 2026-08-30
refactor(client): consolidate relative-time formatters into lib/datetime

### LINKED ISSUE #886: Consolidate four hand-rolled relative-time formatters into lib/datetime.ts
## Problem
Four separate implementations of "5m ago" live in the client:

- `apps/client/src/components/Alerts/AlertDetails/CommentsWall/CommentItem/CommentItem.tsx:40-52`
- `apps/client/src/components/Alerts/AlertDetails/AlertLastCommentSection/AlertLastCommentSection.tsx:13-22` — functionally identical to the one above
- `apps/client/src/pages/Settings.tsx:687` (`formatRelativeTime`, with its own `parseUTCDate` helper alongside it)
- `apps/client/src/pages/MutePolicies.tsx:61-70` — the only one that handles future timestamps ("in 5m")

They have already drifted: future handling exists in exactly one copy.

## Suggested fix
`apps/client/src/lib/datetime.ts` already houses the shared absolute formatters (`formatTime`, `formatShortDateTime`, `formatLongDateTime`, `formatDateTime`, `isSameLocalDay`). Add a `formatRelativeTime` there — covering the future case, since one caller needs it — and replace the four local copies.

Two things to keep in mind:
- Extend the existing tests in `apps/client/src/lib/datetime.test.ts`.
- Settings' `parseUTCDate` exists because the audit-log endpoint still returns SQLite's bare `"YYYY-MM-DD HH:MM:SS"` (UTC with no marker), which `new Date()` reads as local time. Comment timestamps were fixed server-side in #824, but audit logs were not — so preserve that parsing behaviour when you move the code, or the audit times will shift by the viewer's UTC offset.

*(easy)*

AniketR10:
## Issue Reference

closes #886

---

## What Was Changed

added `formatRelativeTime` to `apps/client/src/lib/datetime.ts` and replaced the four hand-rolled "5m ago" implementations in `CommentItem.tsx`, `AlertLastCommentSection.tsx`, `Settings.tsx` (audit log), and `MutePolicies.tsx` with calls to it. extended `datetime.test.ts` with coverage for past/future minutes/hours/days and the week-boundary fallback.

---

## Why Was It Changed

four separate copies of the same relative-time logic had drifted — only `MutePolicies.tsx` handled future timestamps ("in 5m"). consolidating into `lib/datetime.ts` fixes the drift and matches the existing pattern for absolute-time formatters there.

---

## Screenshots

Pasted image (91)
Pasted image (90)

## Summary by CodeRabbit

* **Improvements**
  * Standardized relative timestamps across comments, alerts, mute policies, and audit logs.
  * Improved handling of recent, future, and older dates with configurable relative-time limits and date-time fallbacks.
  * Corrected UTC timestamp interpretation in audit logs.
  * Improved display behavior for missing or invalid timestamps, including preserving available values when dates cannot be parsed.
* **Tests**
  * Added coverage for UTC parsing, relative-time formats, future dates, extended day counts, fallback dates, and invalid values.


AniketR10:
@idanlodzki please review

idanlodzki (commented):
Thanks for this — the diagnosis in #886 was right and the consolidation is clean. Worth saying up front that you got the subtlest part correct: keeping `parseUTCDate` at the audit-log call site rather than passing `log.timestamp` straight into the new formatter. That's the one change here that could have silently broken things, and you handled it.

Heads up that **CI has never actually run on this PR** — as a first-time contributor every real workflow is sitting at `action_required`:

```
CI:                    completed / action_required
Helm Validator:        completed / action_required
Terraform CI workflow: completed / action_required
PR Fields Validation:  completed / action_required
```

The green checks you see are only `label`, `welcome-new-contributor` and `CodeRabbit`. I'll approve the runs. In the meantime I ran it locally with `main` merged in: build 4/4, client tests 250/250, lint 0 errors, `tsc` 0 errors in the touched files. I also grepped for leftovers — the consolidation is complete, no hand-rolled relative-time implementation remains in the client.

To check the refactor didn't change behaviour by accident, I reimplemented all four deleted formatters verbatim and diffed them against `formatRelativeTime` over 29 offsets in both directions. Most of the differences are exactly the de-drifting you intended (`Math.round` → `Math.floor`, future handling generalised, `Invalid Date` no longer reachable). Three came out that I don't think are intended:

### 1. MutePolicies loses its purpose past 7 days

The old `relativeFromNow` had no week cap on purpose — it showed `in 30d`. The new absolute fallback collides with the line rendered directly above it at `MutePolicies.tsx:380`, which is already `formatDateTime(startsAt) → formatDateTime(endsAt)` in that same `formatShortDateTime` shape. So a policy ending in a month now renders:

```
Aug 11, 12:00 → Sep 10, 12:00
ends Sep 10, 12:00              ← previously "ends in 30d"
```

The relative line becomes a verbatim duplicate of the absolute one. Could `formatRelativeTime` take a `maxRelativeDays` (default 7, `Infinity` for this caller), or `MutePolicies.tsx:57` opt out some other way?

### 2. A malformed timestamp now claims the policy is indefinite

`relativeFromNow` passes `'no end'` as the fallback, and the new function returns the caller's fallback for *unparseable* input as well as empty input. The original distinguished the two — `if (!iso) return 'no end'` vs `if (Number.isNaN(ts)) return iso`. So a corrupt `endsAt` now renders `ends no end`, which contradicts the sibling `!s.endsAt → "indefinite"` branch two lines down. A wrong statement about the data is worse than showing the raw string; consider a separate `invalid` fallback, or letting it fall through to the raw value.

### 3. The audit-log wording changed, undisclosed

`5 minutes ago` → `5m ago`, `1 hour ago` → `1h ago`, and the ≥7d fallback goes from `8/4/2026` to `Aug 4, 12:00`. I think the compact form is the right call for consistency, but it isn't mentioned in the PR description and it's the most-read timestamp in Settings — worth calling out explicitly so it's a decision rather than a side effect.

### Two smaller things

- `formatRelativeTime` is the only new lint warning (407 → 408): complexity 11 against the repo max of 10. Collapsing the six `future ? … : …` ternaries into one helper — `const rel = (v: string) => (future ? \`in ${v}\` : \`${v} ago\`)` — takes it back under.
- Since nothing pins the audit-log call site to `parseUTCDate`, a future "simplification" to pass the raw string would silently shift every audit timestamp by the viewer's UTC offset. Measured against a `2m ago` row:

  | TZ | `parseUTCDate` (current) | raw string |
  |---|---|---|
  | UTC | `2m ago` | `2m ago` |
  | Asia/Jerusalem | `2m ago` | **`3h ago`** |
  | America/Los_Angeles | `2m ago` | **`in 6h`** |

  A regression test there would be cheap insurance.

Also needs a `main` merge — the branch is `BEHIND`. Nice work overall; the new test coverage in `datetime.test.ts` is a real improvement on what those four copies had, which was nothing.

AniketR10:
@idanlodzki  have made the requested changes please review

idanlodzki:
Re-reviewed — this addresses all three findings, and the execution is better than what I sketched:

1. **Week cap** — `maxRelativeDays` with MutePolicies passing `Infinity`, and the comment explains *why* that caller opts out. Exactly right.
2. **Malformed timestamps** — `formatRelativeTime(iso, iso || 'no end', Infinity)` is a neat minimal fix: absent → worded fallback, present-but-unparseable → raw text, matching the original byte-for-byte. Verified by re-running my differential harness against the deleted implementation: the only remaining differences are the intended `Math.round`→`Math.floor` de-drift; every cap and edge case now matches.
3. **UTC pinning** — you went further than asked: promoting `parseUTCDate` into `lib/datetime` with the rationale documented and tests asserting both the UTC read and the "2m ago in any zone" behavior is the durable version of what I suggested.

Bonus: the `rel()` helper resolved the complexity-11 lint warning I'd noted, unprompted.

Verified locally on your head merged with current main: build clean, client 257/257, typecheck clean on touched files, no new lint warnings. I've approved the CI runs (the first-contributor gate re-arms on every push — nothing you did wrong).

Nice work. This is ready pending CI.

AniketR10:
@idanlodzki if everything is fine can we merge it now?

PR #892, opened 2026-08-30
fix(server): validate audit pagination and wrap response in envelope

### LINKED ISSUE #881: Audit endpoint: unwrapped response shape + unvalidated pagination
## Problem
`apps/server/src/api/v1/audit/controller.ts:11-19` has two problems in one small handler:

1. **No response envelope.** `return res.json(result)` (line 16) sends the BL result raw, skipping the `{ success, data }` shape every other endpoint uses, and the error path (line 19) returns `{ error: 'Failed to fetch audit logs' }` with no `success: false`. Clients can't branch on `success` for this one endpoint.

2. **Unvalidated pagination.** `parseInt(req.query.page as string) || 1` and `parseInt(req.query.pageSize as string) || 20` (lines 11-12) accept negatives, and nothing caps `pageSize` — `?pageSize=1000000` pulls the whole audit table into memory. `audit/router.ts:6` adds no validation either.

## Suggested fix
Add a small zod schema (`apps/server/src/api/v1/alerts/models.ts` has good neighbours to copy, e.g. `AlertListQueryParamsSchema`) with roughly `page: z.coerce.number().int().min(1)` and `pageSize: z.coerce.number().int().min(1).max(100)`, parse it in the handler, and return `{ success: true, data: result }` / `{ success: false, error: ... }` like the sibling controllers.

*(easy)*

ajayraho:
## Issue Reference
Closes #881

## What Was Changed
- Added `AuditListQueryParamsSchema` (Zod) to validate `page`/`pageSize` on GET /api/v1/audit
- Wrapped the endpoint's response in the standard `{ success, data }` / `{ success: false, error }` envelope
- Updated `Settings.tsx` audit log table to read the new envelope shape
- Extended `audit.test.ts` and fixed `action-audit.test.ts` (both hit this endpoint) for the new response shape, plus added tests for invalid page/pageSize

## Why Was It Changed
The audit endpoint returned raw data instead of the app's standard response envelope, and accepted unvalidated pagination params (negative page, unbounded pageSize). This also required updating the one client call site still reading the raw shape, since the response format change is breaking.

## Screenshots
N/A : backend/API change, no UI changes.

## Additional Context
Tests updated per maintainer guidance on the issue thread. I wasn't able to run the test suite locally due to a native-module build issue in my environment — CI should confirm; happy to fix anything it flags.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated audit log loading to support the standardized success response format.
  - Improved handling of missing or invalid audit log data and API errors.
  - Added validation for pagination values, including valid page ranges and page-size limits.
  - Invalid pagination requests now return clear client errors instead of being silently defaulted.

- **Tests**
  - Expanded coverage for response formatting, default pagination, and invalid pagination parameters.

### DIFF
--- a/apps/client/src/pages/Settings.tsx
+++ b/apps/client/src/pages/Settings.tsx
@@ -710,9 +710,9 @@ const AuditLogTable: React.FC = () => {
 
 		auditApi.getAuditLogs(page, pageSize).then((res) => {
 			if (mounted) {
-				if (res && Array.isArray(res.logs)) {
-					setLogs(res.logs);
-					setTotal(res.total || 0);
+				if (res?.success && res.data && Array.isArray(res.data.logs)) {
+					setLogs(res.data.logs);
+					setTotal(res.data.total || 0);
 					setError(null);
 				} else {
 					setError(res?.error || 'Failed to fetch audit logs');
--- a/apps/server/src/api/v1/audit/controller.ts
+++ b/apps/server/src/api/v1/audit/controller.ts
@@ -1,6 +1,8 @@
 import { Request, Response } from 'express';
 import { AuditBL } from '../../../bl/audit/audit.bl';
 import { Logger } from '@OpsiMate/shared';
+import { AuditListQueryParamsSchema } from './models';
+import { isZodError } from '../../../utils/isZodError.ts';
 
 const logger = new Logger('api/v1/audit/controller');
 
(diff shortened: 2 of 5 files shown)


idanlodzki:
Reviewed and verified this end to end — nice work, and thanks for taking the client half seriously. The envelope change and the `Settings.tsx` call-site update landing together is exactly right; that's what keeps the audit table from breaking.

Verified independently of CI: server 312/312, client 242/242, client typecheck clean, and no conflict with #890 (which also touches `AuditLogTable` — the two auto-merge). Ran every validation path against a live instance:

```
no params (defaults)          200  success=True  logs=7 total=7
page=1&pageSize=3             200  success=True  logs=3 total=7
page=3&pageSize=3             200  success=True  logs=1 total=7
page=0 / page=-1              400  Validation error
pageSize=0 / 1000000 / 101    400  Validation error
pageSize=100 (at cap)         200  success=True
page=abc / page=1.5           400  Validation error
```

Cap boundary is exact, and 100 is safely above the UI — the Items-per-page selector only offers 5/10/15/20. The `isZodError` import style and the `{ success: false, error: 'Validation error', details: error.issues }` shape are byte-identical to the `alerts` and `ai` controllers, so this slots into the existing convention rather than inventing one.

Worth noting two things this fixes that weren't in the issue: `useAuditLogs.ts` was already written against the envelope, and the MSW playground mock already returned it — so the playground audit table was broken before this and works now.

**I pushed one commit to your branch** (`10989fcd`) rather than send you round again for it.

`should default to page 1 / pageSize 20` wasn't testing the default. It asserted `logs.length <= 20` against a table the preceding tests empty, so it passed on zero rows regardless. I confirmed by mutation — changing the schema default from 20 to 3 left the entire audit suite green, so nothing anywhere pinned it:

```
pageSize: ...max(100).optional().default(3)
→ audit.test.ts + action-audit.test.ts: 16/16 still passing
```

Your comment on the test ("No way to observe page/pageSize directly") is the thing to push back on — there is, you just need more than a page of rows. It now seeds 25, then asserts `total === 25`, exactly 20 returned, and that the slice is page 1 (newest row present, 21st absent). Re-ran under mutation: defaults of both 3 and 50 fail it now.

That's the general habit worth taking from this — when a test can't fail, it isn't coverage. Mutating the value under test is the cheapest way to find out.

No action needed from you; I'll merge once CI is green on the new head.

PR #894, opened 2026-08-31
fix: prevent exposing dashboard error details

### LINKED ISSUE #884: Dashboard handlers return raw error.message in 500 responses
## Problem
`apps/server/src/api/v1/dashboards/controller.ts:39-43`, `65-69`, `85-86` — the save, update and delete handlers put raw exception text into the response:

```ts
const message = error instanceof Error ? error.message : String(error);
return res.status(500).json({ success: false, error: message || 'Failed to save dashboard' });
```

That can hand SQL constraint text or filesystem paths to any client.

The tag handlers further down the same file (~135, ~158) already do it the right way: they use `message` only to detect a `'not found'` case for the 404, and return a fixed `'Internal server error'` string for real 500s.

## Suggested fix
Keep logging the error (already happening) and return the generic message on the 500 path in all three handlers, matching the tag handlers in the same file.

*(easy)*

yjrathod:
## Issue Reference
Fixes #884

---

## What Was Changed
Updated dashboard create, update, and delete handlers to return a generic Internal server error

---

## Why Was It Changed
Returning such raw exception to the client, can expose internal server details.

## Summary by CodeRabbit

- **Bug Fixes**
  - Dashboard creation, updates, and deletions now return a clear 404 response when the dashboard cannot be found.
  - Other unexpected errors now return a consistent 500 “Internal server error” response without exposing internal error details.

### DIFF
--- a/apps/server/src/api/v1/dashboards/controller.ts
+++ b/apps/server/src/api/v1/dashboards/controller.ts
@@ -36,11 +36,11 @@ export class DashboardController {
 				return res.status(400).json({ success: false, error: 'Validation error', details: error.issues });
 			}
 			logger.error('Error saving dashboard:', error);
-			const message = error instanceof Error ? error.message : String(error);
-			return res.status(500).json({
-				success: false,
-				error: message || 'Failed to save dashboard',
-			});
+			const message = error instanceof Error ? error.message : 'Internal server error';
+			if (message.includes('not found')) {
+				return res.status(404).json({ success: false, error: 'Dashboard not found' });
+			}
+			return res.status(500).json({ success: false, error: 'Internal server error' });
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


yjrathod:
by the way @idanlodzki, i saw tag handler also returning raw exception message in 404 response, that is also i suppose should to be changed

idanlodzki:
Reviewed and verified live. The fix does exactly what #884 asked, and your instinct to return the fixed string in the 404 (rather than echoing `message` like the tag handlers do) made this better than the pattern it copied.

Probed all three paths against a live instance running this branch:

```
PUT /dashboards/999999            → 404 {"error":"Dashboard not found"}
DELETE (already deleted)          → 404 via the pre-existing boolean path
POST with audit_logs table broken → 500 {"error":"Internal server error"}
```

That last one is the case this PR exists for — before it, the client received the raw `no such table: audit_logs`. Full server suite 312/312, and the client only displays `response.error` verbatim (no matching on text or status), so nothing breaks downstream.

**On your tag-handler observation — you're right.** Both `addTagToDashboardHandler` and `removeTagFromDashboardHandler` do:

```ts
if (message.includes('not found')) {
    return res.status(404).json({ success: false, error: message });
}
```

The leak there is milder than the one you just fixed (that `message` is BL-thrown text like `Tag with ID 3 not found`, not a deep exception), but it's inconsistent with what this PR establishes, and any *unexpected* error whose text happens to contain 'not found' would be echoed raw. If you want to fix it, I'd take it as a follow-up PR rather than growing this one — and it's the natural place to also address CodeRabbit's point here: `message.includes('not found')` is fragile everywhere it appears. A tiny typed error (`class NotFoundError extends Error`) thrown from the BL and checked with `instanceof` in the handlers would fix the string-matching in the dashboard *and* tag handlers at once, and make the dead 404 branches in create/delete (create has no legitimate not-found; delete signals it via the boolean before the catch) fall away naturally.

To be clear on CodeRabbit's finding: I tested its scenario, and the realistic audit failure (`no such table`) doesn't contain 'not found', so it routes to 500 correctly today — it's a robustness cleanup, not a live bug. Not blocking this PR on it.

PR #900, opened 2026-09-05
Fix for Issue #772

### LINKED ISSUE #772: Extract duplicated requireAdmin checks into a shared middleware
## Problem
The same six-line private `requireAdmin` method is copy-pasted verbatim in three controllers:
- `apps/server/src/api/v1/alerts/controller.ts` ~45-51
- `apps/server/src/api/v1/oncall/controller.ts` ~14-20
- `apps/server/src/api/v1/retention/controller.ts` ~18-24

and `apps/server/src/api/v1/users/controller.ts` inlines the same check seven more times (lines ~46, 65, 111, 203, 248).

## Suggested fix
Extract a `requireAdmin` Express middleware next to `authenticateJWT` in `apps/server/src/middleware/auth.ts` and apply it at the router level for admin-only routes.

aryansinha1908:
## Issue Reference

Closes #772

---

## What Was Changed

Extracted the duplicated admin authorization check into a single `requireAdmin`
Express
middleware in `apps/server/src/middleware/auth.ts`, next to `authenticateJWT`, and
applied
it at the router level.

**Removed 18 copies of the same check across 5 controllers:**

| Controller | Removed |
|---|---|
| `alerts/controller.ts` | private helper + 2 call sites |
| `oncall/controller.ts` | private helper + 4 call sites |
| `retention/controller.ts` | private helper + 4 call sites |
| `ai/controller.ts` | private helper + 3 call sites |
| `users/controller.ts` | 5 longhand `if (!req.user \|\| req.user.role !==
Role.Admin)` blocks |

**Applied in the routers:**

- `retention/router.ts` — `router.use(requireAdmin)`; all four routes are
admin-only.
- `alerts`, `oncall`, `ai`, `users` — per-route, since each of these routers mixes
admin and open endpoints (`GET /oncall/teams`, `GET /ai/status`, `POST /ai/filter`,
  `GET /users`, `/users/profile`).

**Incidental cleanups the refactor required:** dropped four now-unused `Role`
imports,
renamed three handler `req` params to `_req` where the guard was their only
consumer, and moved the "reading the schedule is open, changing it is admins only" rationale
comment from the deleted oncall helper onto the router where the split is now expressed.

**New tests:** `apps/server/tests/requireAdmin.test.ts` — 12 tests over all 18
gated routes.

Net **-94 / +51** lines in `src/`.

---

## Why Was It Changed

The same six-line check was maintained in five places, and it had already drifted.
Two concrete symptoms found while doing this:

1. **`ai/controller.ts` had a fourth copy** that the issue didn't list — the drift
was already spreading beyond the three known sites.
2. **`getAllUsersHandler` had silently degraded.** It checks only `if (!req.user)`
— the role comparison is missing — while still returning the `'Forbidden: Admins
only'` message and sitting under a `// (admin only)` router comment. Five near-identical copies is exactly how a load-bearing check loses a clause without anyone noticing.

Putting the gate in `router.ts` also makes it visible where routes are declared, so a reviewer adding an admin route can see the gate rather than having to read the handler body.

**This PR is behavior-preserving.** Status code, error message and JSON body are identical for every request shape:

| Request | Before | After |
|---|---|---|
| Admin JWT | passes | passes |
| Editor / Operation | 403 `Forbidden: Admins only` | same |
| Viewer, write | 403 `Forbidden: Viewer users cannot edit data` (from
`authenticateJWT`, upstream) | same |
| Viewer, read | 403 `Forbidden: Admins only` | same |
| `x-api-token` caller | 403 (API-token auth sets no `req.user`) | same |
| No credentials | 401 from `authenticateJWT` | same |

Ordering is structurally guaranteed: `requireAdmin` reads `req.user`, and all five routers mount below `router.use(authenticateJWT)` at `v1.ts:65`, which is `createV1Router`'s only mount point.

---
## Additional Context

**`GET /users` is deliberately left ungated.** It looks like the same check but isn't, and `apps/client/src/hooks/queries/users/useUsers.ts` is consumed by `AlertsFilterPanel`, `AlertsSelectionBar`, `AlertOwnerColumn`, `useAlertGrouping`, `useAlertSorting`, `AlertInfoSection`, `CommentsWall`, `AlertLastCommentSection` and `pages/Oncall.tsx` — every authenticated role depends on it for owner and comment-author names. 
Applying `requireAdmin` here would 403 the alerts UI for editors, viewers and operation users. The handler keeps its `!req.user` check (which rejects API-token callers) and both it and the route now carry a comment saying why it's open. **The stale `'Forbidden: Admins only'` message and the `(admin only)` router comment were a copy-paste artifact; the router comment is corrected here, the response body is left untouched since changing it is client-visible.**
If the gate is actually wanted, that's a deliberate breaking change and belongs in its own issue.

**One intentional divergence:** `router.use(requireAdmin)` on the retention router means unmatched paths under `/api/v1/retention/*` now return **403 instead of 404** for non-admins.
Arguably an improvement (it stops leaking route existence), but flagging it as a real change.

**Type narrowing at `users/controller.ts`.** `updateUserPasswordHandler`'s self-reset guard reads `req.user.id` and only compiled because the deleted check narrowed `req.user`. It's now `if (req.user && userId === parseInt(req.user.id, 10))` — eslint's `recommended` config bans non-null assertions, so `req.user!` wasn't an option. `ai/controller.ts` needed no equivalent change: `AiBL.updateConfig(updates, user?: User)` already takes an optional user.

**No boolean helper was exported alongside the middleware.** After this refactor zero call sites need one, and exporting two shapes of the same rule is how the duplication would grow back.

### Verification

- `pnpm --filter @OpsiMate/server test` — **351 passed** (339 pre-existing + 12
new).
- Typecheck clean.
- The new tests were mutation-checked: neutering `requireAdmin` to a bare `next()`
fails 5
  of the 12, so they aren't passing vacuously.
- Pre-existing regressions kept green: `alerts.test.ts:1799` (silence-reset rejects
  non-admins), `aiConfig.test.ts:117` (viewer 403 on `GET /ai/config`),
`auth.test.ts:174`
  (non-admin cannot delete a user).

The new suite asserts the *layering*, not just the status code — 401 before 403 for unauthenticated calls, viewers getting the write-block message on writes vs admins-only on reads, API-token callers rejected, and route-ordering locks (`/users/profile` before `/:id`, the silence-reset gate not leaking onto `/alerts/:alertId`) that would catch a stray `router.use` on a mixed router.

## Summary by CodeRabbit

- **Access Control**
  - Centralized admin authorization for AI configuration and connection testing, alert silence-reset settings, retention management, on-call team changes, and user administration.
  - Non-admin authenticated users can continue accessing designated shared endpoints, including on-call team lists, AI status, user lists, and alert-related data.
  - Unauthorized requests now receive consistent authentication or forbidden responses across protected routes.

- **Tests**
  - Added coverage for protected routes, permitted shared routes, role-based responses, and route-isolation behavior.

### DIFF
--- a/apps/server/src/api/v1/ai/controller.ts
+++ b/apps/server/src/api/v1/ai/controller.ts
@@ -1,5 +1,5 @@
 import { Response } from 'express';
-import { AiFilterQuerySchema, Logger, Role, UpdateAiConfigSchema } from '@OpsiMate/shared';
+import { AiFilterQuerySchema, Logger, UpdateAiConfigSchema } from '@OpsiMate/shared';
 import { AiBL, AiDisabledError, AiValidationError, BedrockCallError } from '../../../bl/ai/ai.bl';
 import { AuthenticatedRequest } from '../../../middleware/auth';
 import { isZodError } from '../../../utils/isZodError.ts';
@@ -34,18 +34,7 @@ export class AiController {
 		return true;
 	}
 
-	// Org-wide configuration holding a credential — admin-only, same gate the
-	// retention and silence-reset settings use.
-	private requireAdmin(req: AuthenticatedRequest, res: Response): boolean {
-		if (!req.user || req.user.role !== Role.Admin) {
-			res.status(403).json({ success: false, error: 'Forbidden: Admins only' });
-			return false;
-		}
-		return true;
-	}
-
-	getConfigHandler = async (req: AuthenticatedRequest, res: Response) => {
-		if (!this.requireAdmin(req, res)) return;
+	getConfigHandler = async (_req: AuthenticatedRequest, res: Response) => {
 		try {
 			const config = await this.aiBL.getConfig();
 			return res.json({ success: true, data: config });
@@ -56,7 +45,6 @@ export class AiController {
 	};
 
 	updateConfigHandler = async (req: AuthenticatedRequest, res: Response) => {
-		if (!this.requireAdmin(req, res)) return;
 		try {
 			const updates = UpdateAiConfigSchema.parse(req.body ?? {});
 			const config = await this.aiBL.updateConfig(updates, req.user);
@@ -111,8 +99,7 @@ export class AiController {
 		}
 	};
 
-	testConnectionHandler = async (req: AuthenticatedRequest, res: Response) => {
-		if (!this.requireAdmin(req, res)) return;
+	testConnectionHandler = async (_req: AuthenticatedRequest, res: Response) => {
 		try {
 			const result = await this.aiBL.testConnection();
 			// Always 200: a failed Bedrock call is a RESULT the user asked for, not a
--- a/apps/server/src/api/v1/ai/router.ts
+++ b/apps/server/src/api/v1/ai/router.ts
@@ -1,14 +1,15 @@
 import { Router } from 'express';
 import { AiController } from './controller';
+import { requireAdmin } from '../../../middleware/auth';
 
 export default function createAiRouter(controller: AiController) {
 	const router = Router();
 
 	router.get('/status', controller.statusHandler);
 	router.post('/filter', controller.filterHandler);
-	router.get('/config', controller.getConfigHandler);
-	router.put('/config', controller.updateConfigHandler);
-	router.post('/test', controller.testConnectionHandler);
+	router.get('/config', requireAdmin, controller.getConfigHandler);
+	router.put('/config', requireAdmin, controller.updateConfigHandler);
+	router.post('/test', requireAdmin, controller.testConnectionHandler);
 
 	return router;
 }
--- a/apps/server/src/api/v1/alerts/controller.ts
+++ b/apps/server/src/api/v1/alerts/controller.ts
@@ -6,7 +6,6 @@ import {
 	AlertStatus,
 	CreateCommentSchema,
 	Logger,
-	Role,
 	UpdateCommentSchema,
 	UpdateSilenceResetSettingsSchema,
 } from '@OpsiMate/shared';
@@ -179,18 +178,7 @@ export class AlertController {
 		return res.type('application/json').send(`{"success":true,"data":{"alerts":${snapshot.json}}}`);
 	}
 
-	// Both silence-reset endpoints are admin-only: this is org-wide configuration, same
-	// gate the retention settings use.
-	private requireAdmin(req: AuthenticatedRequest, res: Response): boolean {
-		if (!req.user || req.user.role !== Role.Admin) {
-			res.status(403).json({ success: false, error: 'Forbidden: Admins only' });
-			return false;
-		}
-		return true;
-	}
-
-	async getSilenceResetSettings(req: AuthenticatedRequest, res: Response) {
-		if (!this.requireAdmin(req, res)) return;
+	async getSilenceResetSettings(_req: AuthenticatedRequest, res: Response) {
 		try {
 			const settings = await this.alertBL.getSilenceResetSettings();
 			return res.json({ success: true, data: settings });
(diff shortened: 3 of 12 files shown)


idanlodzki:
Reviewed this carefully — an auth refactor is exactly where a subtle protection change hides, so I diffed the full protection map against main: **all 18 previously-gated call sites map 1:1 to the new router-level gates, and nothing gained or lost a gate.** The judgment calls are right too — `GET /users` left open with the reason documented (the old "(admin only)" comment was aspirational; the alerts UI needs it for every role), `retention` as a router-wide `use`, mixed routers per-route. And `requireAdmin.test.ts` is the best part of the PR: enumerating every gated route and pinning the layering (401 before 403, API-token rejection) is what makes this refactor safe to merge and *stay* safe.

Two small things to fix before merge, both from CodeRabbit and both quick:

1. **`users/router.ts` route comments drifted from the paths** — "PATCH /users/password" sits on `/:id/reset-password` and "PATCH /user/:id" should be `/users/:id`. On a security-sensitive router, comments that misname endpoints are worse than none; please align them.
2. **`requireAdmin.test.ts:42`** — the inline `Record<string, (p: string) => Test>` cast violates the repo's named-interfaces rule (no anonymous inline object types, tests included; CodeRabbit enforces it from org settings so it only surfaces in review). Declare an interface for the supertest verbs and type `method` as its key, e.g.:

```ts
interface HttpMethodCaller {
	get: (path: string) => Test;
	post: (path: string) => Test;
	put: (path: string) => Test;
	patch: (path: string) => Test;
	delete: (path: string) => Test;
}
const send = (method: keyof HttpMethodCaller, path: string) => (app as unknown as HttpMethodCaller)[method](path);
```
(and tighten `ADMIN_ROUTES` to `[keyof HttpMethodCaller, string][]` so the verbs stay checked).

Push those two and this is ready. CI is already green — the approval gate re-arms on every push from a first-time fork, so ping here if your next push sits at "action required" and I'll approve it. Genuinely strong first contribution.

idanlodzki:
Checking in — the two small fixes from my review (route comments in `users/router.ts`, and the named interface for the supertest verbs in `requireAdmin.test.ts`) are all that stand between this and merge. The rest of the refactor is verified and solid. Are you still able to push them? If you're short on time, say so and I'll apply them on top of your branch with you as the author of the PR.

aryansinha1908:
@idanlodzki Sorry for the late response, I will push the fixes soon.

idanlodzki:
Hi again — no pressure at all, I know life gets busy. Your refactor is verified and in really good shape; the only things left are the two tiny items above (the route comments and the named interface in the test), and honestly they're a five-minute job.

To save you the round-trip, I'm happy to just apply those two on top of your branch and merge — you'd stay the author, and it would land as your contribution. If you'd rather make the changes yourself, that's great too, just say so and I'll wait. Otherwise I'll go ahead in a few days so your work gets in. Thanks again for taking this on — it's a genuinely good first contribution.

aryansinha1908:
I am so sorry for these late changes. I have alot of things going on in my college. I hope the changes are satisfactory.

idanlodzki:
No apology needed — college comes first, and the changes are exactly right. Both items are done: the route comments now name the real paths, and the supertest verbs go through a named `HttpMethodCaller` interface. You went further and added two tests pinning that `/users/profile` and `/users/role` still match before the gated `/:id` — that's precisely the regression an auth refactor needs, thank you.

Verified on top of current main: 386/386 server tests, lint at CI strictness, Prettier clean. I've approved the CI run (first-time forks are gated on every push). This is ready; it merges once the maintainer signs off, which should be soon. Really good first contribution — and if you like the project, a ⭐ on the repo helps us a lot.

idanlodzki:
Merging — thank you! 🎉

PR #902, opened 2026-09-06
fix : Add keyboard navigation to dashboard search

### LINKED ISSUE #883: Dashboard search suggestions can't be selected with the keyboard
## Problem
`apps/client/src/components/Alerts/DashboardHeader.tsx:136-161` — the dashboard search dropdown is mouse-only:

- suggestions are chosen with `onMouseDown` (line 150): no arrow-key navigation, no Enter to select
- no `role="combobox"` / `role="listbox"` / `role="option"`, no `aria-expanded` or `aria-activedescendant`, so screen readers get nothing
- the list is dismissed by `onBlur={() => setTimeout(() => setIsSearchFocused(false), 200)}` (line 140) — a magic delay racing the click

A keyboard user can type a query but cannot reach the results.

## Suggested fix
`apps/client/src/components/shared/MatcherGroupsEditor/AutocompleteInput.tsx` already solves exactly this problem in this codebase: a styled listbox with ArrowUp/ArrowDown/Enter/Escape handling, selection committed on `onMouseDown` so it fires before blur, and click-away handled by a document listener instead of a timeout. Follow that pattern here, or reuse the component if it fits.

*(easy)*

AnuragR3k:
## Issue Reference

Closes #883

---

## What Was Changed

- Added `ArrowDown` and `ArrowUp` keyboard navigation for dashboard search suggestions.
- Added `Enter` to select the highlighted dashboard.
- Added `Escape` to close the search suggestions.
- Added active suggestion highlighting during keyboard navigation.
- Added accessibility attributes for the search combobox and suggestion list.
- Replaced the delayed `onBlur` handling with click-away detection.

---

## Why Was It Changed

Dashboard search suggestions could previously only be selected using the mouse. This prevented keyboard users from navigating and selecting dashboard suggestions.

This change makes the search dropdown keyboard accessible while keeping the existing mouse interaction intact.

---

## Screenshots

https://github.com/user-attachments/assets/e22d77de-d6b3-4a4a-aa68-b6e30af0a211

---

- Verified `ArrowUp` / `ArrowDown`, `Enter`, and `Escape` behavior locally.
- Tested with multiple dashboard suggestions.
- Ran `pnpm run check` successfully.

## Summary by CodeRabbit

- **New Features**
  - Improved dashboard search accessibility by exposing the currently active result to assistive technologies.
  - Restored scrolling for longer dashboard search result lists.

- **Bug Fixes**
  - Prevented stale or invalid selections when dashboard search results change.
  - Enter-key selection now safely handles cases where the selected dashboard is no longer available.
  - Suggestions now close correctly when focus moves outside the search area.

### DIFF
--- a/apps/client/src/components/Alerts/DashboardHeader.tsx
+++ b/apps/client/src/components/Alerts/DashboardHeader.tsx
@@ -181,7 +181,16 @@ export const DashboardHeader = ({
 			</div>
 
 			{/* Search (grows to fill the row) */}
-			<div ref={searchContainerRef} className="relative flex-1 min-w-0">
+			<div
+				ref={searchContainerRef}
+				className="relative flex-1 min-w-0"
+				onBlur={(e) => {
+					if (!searchContainerRef.current?.contains(e.relatedTarget as Node | null)) {
+						setIsSearchFocused(false);
+						setActiveIndex(-1);
+					}
+				}}
+			>
 				<div className="flex items-center h-8 rounded-md border bg-background px-3 focus-within:ring-1 focus-within:ring-ring">
 					<Search className="mr-2 h-4 w-4 shrink-0 opacity-50 text-foreground" />
 					<input
--- a/apps/client/src/components/Alerts/DashboardHeader.tsx
+++ b/apps/client/src/components/Alerts/DashboardHeader.tsx
@@ -57,19 +57,23 @@ export const DashboardHeader = ({
 			inputRef.current.select();
 		}
 	}, [isEditingName]);
-	
+
+	useEffect(() => {
+		setActiveIndex(-1);
+	}, [filteredDashboards]);
+
 	useEffect(() => {
 		if (!isSearchFocused) return;
-	
+
 		const handleClickAway = (e: MouseEvent) => {
 			if (!searchContainerRef.current?.contains(e.target as Node)) {
 				setIsSearchFocused(false);
 				setActiveIndex(-1);
 			}
 		};
-	
+
 		document.addEventListener('mousedown', handleClickAway);
-	
+
 		return () => {
 			document.removeEventListener('mousedown', handleClickAway);
 		};
@@ -107,11 +111,12 @@ export const DashboardHeader = ({
 
 			const dashboard = filteredDashboards[activeIndex];
 
-			onDashboardSelect?.(dashboard);
-			setSearchQuery('');
-			setIsSearchFocused(false);
-			setActiveIndex(-1);
-			searchInputRef.current?.blur();
+			if (dashboard) {
+				onDashboardSelect?.(dashboard);
+				setSearchQuery('');
+				setIsSearchFocused(false);
+				setActiveIndex(-1);
+			}
 		}
 		if (e.key === 'Escape') {
 			e.preventDefault();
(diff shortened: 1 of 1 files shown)


AnuragR3k:
CI is currently failing on an existing lint error in `apps/client/src/components/AddUserModal.tsx:58` (`error` is defined but never used). This file is unrelated to the changes in this PR, and `pnpm run check` passed locally before opening the PR.

idanlodzki:
Thanks for this — the keyboard model is right (arrows wrap, Enter selects, Escape dismisses, `role="combobox"`/`listbox`/`option` semantics), and the click-away listener is a genuinely better dismissal than the old 200ms blur timeout. The recording in the description helps a lot. A few things before it can merge:

### CI is failing — prettier, not logic
The `Format apps/client` step runs `prettier --check`, and the new `useEffect` block has whitespace-only lines carrying tabs (the blank lines around `handleClickAway`). `pnpm run check` doesn't cover this — run `pnpm --filter @OpsiMate/client format-fix` and it'll rewrite the file. (For what it's worth, I tripped over the exact same thing on #898 last week.)

### One regression CodeRabbit didn't flag
The `<ul>` lost `className="max-h-[300px] overflow-y-auto py-1"` when you added `id`/`role` to it. That capped the dropdown's height and made it scroll — with many dashboards, the list now grows unbounded off the bottom of the viewport. Please restore those classes alongside the new attributes.

### CodeRabbit's three findings — all valid, please address
1. **Stale `activeIndex` can crash Enter** (its Major finding, and it's real): `activeIndex` resets on query change, but not when `dashboards` itself changes while the list is open — if the list shrinks, `filteredDashboards[activeIndex]` is `undefined` and `onDashboardSelect` throws on `.id`. Cheapest fix: reset it whenever the result set changes —
   ```ts
   useEffect(() => setActiveIndex(-1), [filteredDashboards]);
   ```
   and guard the Enter path with `if (dashboard)` anyway.
2. **`aria-activedescendant`**: focus stays on the input during arrow navigation, so screen readers can't tell which option is active from `aria-selected` alone. Give each `<li>` a stable `id` (e.g. `dashboard-option-${dashboard.id}`) and set `aria-activedescendant` on the input to the active one.
3. **Tab leaves the list open**: `onBlur={() => {}}` plus a `mousedown`-only click-away means tabbing out leaves the listbox visible with `aria-expanded` true. Handle it in the container's `onBlur` (check `e.relatedTarget` against `searchContainerRef`) rather than a no-op on the input.

Two small notes: the PR title has the description doubled ("fix : Add keyboard navigation to dashboard searchfix: add keyboard navigation…") — worth tidying since it becomes the squash commit — and after your next push the first-contributor CI gate re-arms; ping here and I'll approve it.

Good bones; these are all small.

AnuragR3k:
Addressed the requested changes and pushed the fixes.

- Fixed the Prettier formatting issue
- Restored the dropdown max-height and scrolling
- Guarded against stale activeIndex values
- Added aria-activedescendant and stable option IDs
- Fixed Tab/blur dismissal behavior

CI is re-running. Thanks for the review!

idanlodzki:
Fast turnaround — four of the five are exactly right, including the `if (dashboard)` guard and the `[filteredDashboards]` reset (that was the crash), the restored scroll classes, and `aria-activedescendant` with stable option ids. Prettier is clean now too.

One is still open: **Tab still leaves the list showing.** The fix removed the `onBlur={() => {}}` no-op, but nothing replaced it — the only dismissal paths are `mousedown` (click-away) and Escape, and Tab triggers neither. Try it: focus the search with results showing, press Tab → focus moves to the next control, the listbox stays visible with `aria-expanded` still true.

The fix is a `onBlur` on the **container** (`searchContainerRef`'s div), not the input, so moving focus *between* the input and an option doesn't count as leaving:

```tsx
<div
	ref={searchContainerRef}
	className="relative flex-1 min-w-0"
	onBlur={(e) => {
		// Focus moved outside the search area (Tab, or click on another control).
		if (!searchContainerRef.current?.contains(e.relatedTarget as Node | null)) {
			setIsSearchFocused(false);
			setActiveIndex(-1);
		}
	}}
>
```

React's `onBlur` bubbles (unlike native `blur`), so the container sees the input losing focus, and `relatedTarget` is where focus went. Once that's in and CI is green this is ready to merge.

AnuragR3k:
Addressed the remaining Tab/blur issue and pushed the fix 

Tab now moves focus outside the search area and the suggestions close. CI/check is green locally with 0 errors.

Thanks for the review! and sry for this mistakes :(

https://github.com/user-attachments/assets/4aeb900f-39ad-41b3-82fa-a99e5a7afdb8

PR #929, opened 2026-09-11
[FIX]: Add regression coverage for client API error messages

### LINKED ISSUE #924: Add unit tests for lib/errorMapper
## Problem
`apps/client/src/lib/errorMapper.ts` turns API error responses into user-facing messages for every form in the app, and has no tests.

## What to test
New file `apps/client/src/test/errorMapper.test.ts`:
- `mapValidationErrors`: field-specific detail beats the generic message; multiple fields map independently
- `mapApiError`: the HTTP-status-ish messages (401/403/404/500 wording) and the passthrough for unknown strings
- `getFieldError`: returns the field's message, `null` when the field has no detail
- `isValidationError` type guard: true for `{ success: false, error: 'Validation error', details: [...] }`; false for `null`, `undefined`, `{}`, a success response, and a non-validation error

*(easy — pure functions; the guard already handles null via optional chaining, so these tests pin that it stays that way)*

sridharkalaibala:
## Issue Reference

Closes #924

## What Was Changed

Added 33 Vitest cases for the four public helpers in `errorMapper.ts`: validation-message precedence and fallbacks, independent and nested fields, HTTP and authentication/network messages, unmapped strings, field lookup, and validation-response detection.

## Why Was It Changed

These helpers provide error messages throughout the application's forms. The tests protect the existing user-facing messages and ensure unrelated or incomplete responses are not treated as validation errors.

## Screenshots

Not applicable; this change adds tests only.

Validated on Windows with Node 24 and pnpm 9.12.0:

- Full client suite: 35 files, 294 tests passed (33 new tests).
- Client production build and both workspace dependency builds passed.
- ESLint and Prettier checks passed for the new test file.
- Direct client `tsc` output was compared with the committed type-error baseline, normalizing the local repository path: no new diagnostics. Existing baseline errors remain.

Prepared with AI assistance; the tests were executed locally and the change was reviewed against the issue's requested cases.

## Summary by CodeRabbit

* **Tests**
  * Added automated coverage for validation error mapping, API error translation, field-level error lookup, nested fields, and validation error detection.
  * Added fallback coverage for unrecognized error codes.

### DIFF
--- a/apps/client/src/test/errorMapper.test.ts
+++ b/apps/client/src/test/errorMapper.test.ts
@@ -1,0 +1,135 @@
+import { describe, expect, it } from 'vitest';
+import {
+	getFieldError,
+	isValidationError,
+	mapApiError,
+	mapValidationErrors,
+	type ValidationErrorDetail,
+	type ValidationErrorResponse,
+} from '@/lib/errorMapper';
+
+const emailError: ValidationErrorDetail = {
+	code: 'invalid_string',
+	type: 'string',
+	message: 'Invalid',
+	path: ['email'],
+};
+
+function validationResponse(details: ValidationErrorDetail[]): ValidationErrorResponse {
+	return { success: false, error: 'Validation error', details };
+}
+
+describe('mapValidationErrors', () => {
+	it('prefers field-specific messages over generic and server messages', () => {
+		expect(mapValidationErrors(validationResponse([emailError]))).toEqual({
+			email: 'Please enter a valid email address',
+		});
+	});
+
+	it('maps multiple fields independently, including nested field paths', () => {
+		const response = validationResponse([
+			emailError,
+			{ code: 'too_small', type: 'string', message: 'Too short', path: ['password'] },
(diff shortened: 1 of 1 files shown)


idanlodzki:
Reviewed and verified — this is exactly what #924 asked for, and then some. You covered every case in the issue plus several I hadn't listed: nested field paths, the `Validation error: …` prefixed variant, and `details: {}` (present but not an array). The `it.each` tables for the message maps are the right shape — the matrix *is* the documentation.

I don't take "tests pass" at face value on a test-only PR, so I mutated the source to see if the suite actually bites:
- removed the null-guard in `isValidationError` → exactly the `null` / `undefined` cases fail
- changed the 401 message → exactly that one case fails

That's the property that makes tests worth having, and it holds. Client suite 294/294, prettier clean, lint clean at `--max-warnings=0`. CI was waiting at the first-contributor approval gate — nothing you did; I've approved the runs and will merge when it's green.

One small ask, since you clearly took the time to understand this codebase: if OpsiMate is useful to you, a ⭐ on the repo genuinely helps — it's how other people find it. Welcome aboard, and thanks for a careful first contribution.

PR #931, opened 2026-09-11
[FIX]: Reject failed typecheck commands without diagnostics

sridharkalaibala:
## Issue Reference

Found while validating the client locally; no existing issue found for failed compiler commands passing this gate.

## What Was Changed

Require at least one recognized TypeScript diagnostic before accepting a failed `tsc` command with exit code 1 or 2. Command failures without diagnostics now exit 2 before comparing or rewriting the baseline.

## Why Was It Changed

Package-manager failures can return the same exit codes as TypeScript. For example, `pnpm exec tsc` failing with `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL` was normalized to an empty diagnostic list and reported as a passing type check. Running `typecheck:update` in that situation could erase the committed baseline. Successful compiler runs with no diagnostics and genuine baselined TypeScript errors keep their existing behavior.

## Screenshots

Not applicable: this changes the command-line type-check gate, with no UI changes.

- Added nine Vitest tests loading the actual gate script with mocked command output, filesystem access, and process exits. Four regression cases failed before the fix and all nine pass afterward.
- Coverage includes command errors with exit 1/2, silent failures, preserving the baseline on a failed update, accepting known diagnostics, rejecting a new global diagnostic, successful runs, and intentionally clearing the baseline after success.
- Full client suite: **303 tests across 36 files pass**. Changed files pass ESLint, Prettier, JavaScript syntax checks, and `git diff --check`.
- The tests simulate compiler outcomes; they do not install or invoke another TypeScript compiler. No application code, dependencies, or baseline entries changed.
- AI assistance: developed and validated with OpenAI Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type-check validation so command and shell failures are no longer treated as successful runs without diagnostics.
  - Applied the same validation when updating the diagnostic baseline.

- **Tests**
  - Added coverage for successful checks, command failures, missing output, existing diagnostics, newly detected diagnostics, and baseline updates.

### DIFF
--- a/apps/client/scripts/check-types.mjs
+++ b/apps/client/scripts/check-types.mjs
@@ -20,9 +20,9 @@ const baselineFile = join(clientDir, 'type-errors-baseline.txt');
 // would never match across environments.
 const repoRoot = resolve(clientDir, '..', '..');
 
-// tsc exits 1 (or 2) when it reports diagnostics — expected. Any OTHER failure means
-// tsc never actually ran (binary missing, bad config path), which must fail the gate
-// loudly instead of looking like "no errors".
+// tsc exits 1 (or 2) when it reports diagnostics — expected. Package-manager and
+// shell failures can use the same exit codes, so require actual diagnostics too.
+// Otherwise a failed invocation would look like "no errors", even in --update mode.
 const runTsc = () => {
 	try {
 		execSync('pnpm exec tsc -p tsconfig.app.json --noEmit', {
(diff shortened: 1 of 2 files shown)


idanlodzki:
This is a genuinely sharp find, and not something I filed — thank you for going looking.

The gate's own comment promised that a failed invocation "must fail loudly instead of looking like no errors," but a `pnpm exec tsc` failure exits 1 with zero TS diagnostics, `normalize()` turns that into an empty set, and the gate passes green. And you're right that `--update` in that state would **erase the baseline** — that's the dangerous half. The fix — require at least one recognized diagnostic before accepting exit 1/2 — is exactly right and minimal.

Verified: 9/9 tests, and critically the **real gate still passes on current main** (`No new client type errors (25 known baseline error(s))` — it even noticed one baseline error no longer occurs). Prettier and lint clean at `--max-warnings=0`. Mocking `execSync`/fs/`process.exit` to test the actual script rather than a copy of its logic is the right call.

Two small notes, neither blocking: the test file is `.js` in a TS repo — fine for a script test, just unusual; and the 25-vs-baseline delta message suggests running `typecheck:update` on main soon, unrelated to you. Merging when CI is green. You've now landed two quality PRs in a day — if you haven't yet, a ⭐ on the repo would be appreciated.

PR #933, opened 2026-09-11
[FIX]: Standardize client on shared Role enum

### LINKED ISSUE #917: Role enum is declared three times — standardize on the shared export
## Problem
The `Role` enum exists in three places with identical members:

- `packages/shared/src/types.ts:22` — the canonical, exported one
- `apps/client/src/types/index.ts:2` — a client copy, exported
- `apps/client/src/lib/permissions.ts:5` — another copy, not even exported

Different client modules import different copies. They happen to agree today; the first time one is edited, they won't — and TypeScript will treat `Role.Admin` from one module as incompatible with `Role.Admin` from another.

## Suggested fix
Delete the two client copies and import `Role` from `@OpsiMate/shared` everywhere (grep `Role\b` imports in `apps/client/src`). `permissions.ts` already imports other things from shared, so it's a one-line change there. Run `pnpm --filter @OpsiMate/client typecheck` after — any import that still points at the deleted copy will fail loudly.

*(easy — mechanical, but it touches every file that imports `Role`, so it's good practice for a careful multi-file refactor)*

JavierQuinan:
## Issue Reference

Closes #917

---

## What Was Changed

- Removed the duplicated `Role` enum from `apps/client/src/types/index.ts`.
- Re-exported the canonical `Role` enum from `@OpsiMate/shared` to preserve existing client imports.
- Removed the local `Role` enum from `apps/client/src/lib/permissions.ts`.
- Updated `permissions.ts` to use the shared `Role` enum.
- Updated `AddUserModal.tsx` to import `Role` directly from `@OpsiMate/shared`.

---

## Why Was It Changed

The `Role` enum was defined in multiple locations with the same values.

Keeping separate enum declarations creates a risk of the definitions diverging over time and can also cause TypeScript incompatibilities between values originating from different enum declarations.

This change makes `packages/shared/src/types.ts` the single source of truth for `Role` while keeping the existing client barrel export compatible.

---

## Screenshots

Not applicable. This change does not modify the UI.

---

## Additional Context

This is intentionally a small refactor.

No role values, permission rules, API behavior, or user-facing functionality were changed.

The branch only updates the client-side references needed to use the shared `Role` enum.

## Summary by CodeRabbit

- **Refactor**
  - Standardized role definitions across the client to improve consistency in user and permission handling.
  - Updated role assignments and permission checks to use the same shared definitions.
  - Removed the separate client-side role definition while preserving existing roles and permission behavior.
  - No changes to the available user experience or role-based capabilities.

### DIFF
--- a/apps/client/src/components/AddUserModal.tsx
+++ b/apps/client/src/components/AddUserModal.tsx
@@ -5,7 +5,8 @@ import { Label } from './ui/label';
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from './ui/select';
 import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogHeader, DialogTitle } from './ui/dialog';
 import { apiRequest } from '../lib/api';
-import { User, Role } from '../types';
+import { Role } from '@OpsiMate/shared';
+import { User } from '../types';
 import { ErrorAlert } from './ErrorAlert';
 import { useFormErrors } from '../hooks/useFormErrors';
 
--- a/apps/client/src/lib/permissions.ts
+++ b/apps/client/src/lib/permissions.ts
@@ -1,13 +1,6 @@
+import { Role } from '@OpsiMate/shared';
 import { getUserRole } from './auth';
 import { isPlaygroundMode } from './playground';
-
-// Local role definitions for client-side use
-enum Role {
-	Admin = 'admin',
-	Editor = 'editor',
-	Viewer = 'viewer',
-	Operation = 'operation',
-}
 
 export type Permission = 'create' | 'edit' | 'delete' | 'view' | 'operate';
 
--- a/apps/client/src/types/index.ts
+++ b/apps/client/src/types/index.ts
@@ -1,10 +1,6 @@
-// Client-side type definitions
-export enum Role {
-	Admin = 'admin',
-	Editor = 'editor',
-	Viewer = 'viewer',
-	Operation = 'operation',
-}
+import { Role } from '@OpsiMate/shared';
+
+export { Role } from '@OpsiMate/shared';
 
 export interface User {
 	id: number;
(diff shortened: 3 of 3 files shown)


idanlodzki:
Right approach — turning `types/index.ts` into a re-export (`export { Role } from '@OpsiMate/shared'`) instead of deleting the symbol means nothing importing `Role` from `../types` breaks, and I confirmed the shared enum has the same four members (`Role.Operation` is used in 7 places; all safe). Tests 261/261, prettier clean.

One real problem, and it's the kind that hides: **a re-export doesn't bring the name into scope.** `types/index.ts` line 7 uses `Role` itself (`role: Role` on `User`), and with only `export { Role } from …` there's no local binding, so:

```
src/types/index.ts(7,8): error TS2304: Cannot find name 'Role'.
```

Vite and Vitest don't catch it (esbuild strips types without checking), which is why everything looked green locally — but CI runs `pnpm --filter @OpsiMate/client typecheck`, which gates against a baseline of known errors, and this one is **new**, so CI will fail. Fix is one line — import it as well as re-export it:

```ts
import { Role } from '@OpsiMate/shared';
export { Role };
```

Run `pnpm --filter @OpsiMate/client typecheck` before pushing and it should say "No new client type errors". After your push the first-contributor CI gate re-arms; ping here and I'll approve it. Close to done — nice refactor otherwise.

JavierQuinan:
Fixed in 15b043b9: Role is now imported into local scope and re-exported separately, so User.role: Role type-checks correctly. The first-contributor workflows have re-armed and are ready for approval.

idanlodzki:
Fix confirmed — `import { Role } from '@OpsiMate/shared'; export { Role };` is exactly right, and the real CI gate now reports `No new client type errors`. Tests 261/261. Enum parity with the shared copy verified (four members; `Role.Operation`'s seven call sites are all safe). Merging when CI is green — thanks for the quick turnaround, and if OpsiMate is useful to you, a ⭐ helps others find it.

Issue #927, opened 2026-09-11
Add unit tests for lib/permissions (role × permission matrix)


idanlodzki:
## Problem
`apps/client/src/lib/permissions.ts` gates every write control in the UI by role, and has no tests. A wrong entry means a viewer sees a delete button (or an admin doesn't).

## What to test
New file `apps/client/src/test/permissions.test.ts`. This is the good "learn `vi.mock`" task: mock `getUserRole` and `isPlaygroundMode`, then assert the full matrix — for each of admin / editor / viewer, which permissions return true and which false — plus that playground mode grants admin-level permissions regardless of the stored role.

Heads up: #917 removes this file's local `Role` enum in favour of the shared one — either PR can land first, just import `Role` from `@OpsiMate/shared` in the test.

*(easy — the matrix itself is the documentation)*

idanlodzki:
Correction to the issue text: playground mode is **read-only**, not admin-level — `hasPermission` returns false for every permission except `view` when `isPlaygroundMode()`, regardless of role. #934 tests the real behavior; the issue's wording was wrong.

idanlodzki:
Closed by #934 (the PR omitted the 'Closes' keyword).

PR #934, opened 2026-09-11
[FIX]: Cover role permissions and playground restrictions

sridharkalaibala:
Add a table-driven permission suite covering all four shared roles plus a missing role, every permission and convenience helper, feature-management checks, and the read-only playground override.

Addresses #927. The issue describes playground mode as granting admin permissions, but current `getCurrentUser()` only grants the admin identity: `hasPermission()` explicitly rejects non-view actions when `isReadOnlyMode()` is true, and both user/integration management helpers do the same. These tests preserve that deliberate write restriction. No permission policy changes are included.

Normal-mode expectations:

| Role | Create | Edit | Delete | View | Operate | Manage users/integrations | Manage services/tags |
|---|---|---|---|---|---|---|---|
| Admin | yes | yes | yes | yes | yes | yes | yes |
| Editor | yes | yes | no | yes | yes | no | yes |
| Viewer | no | no | no | yes | no | no | no |
| Operation | no | no | no | yes | yes | no | no |
| Missing | no | no | no | no | no | no | no |

Playground mode blocks all write/management actions for every mocked role, including admin. The test imports `Role` from `@OpsiMate/shared`, so it is compatible with the enum consolidation in #933.

Validation:
- 62 focused tests passed, mocking `getUserRole` and `isPlaygroundMode` as requested.
- Mutation check: temporarily granting viewers every permission caused five tests to fail. Restored production source afterward; the PR contains only the new test file.

AI assistance: OpenAI Codex helped implement and validate this suite.

Additional validation: full client suite passed (378 tests in 39 files); changed-file ESLint passed with `--max-warnings=0`; Prettier and `git diff --check` passed. Ran installed Vitest/ESLint binaries directly: the default pnpm 11 command attempted installation and stopped at ignored-build policy checks, so no dependency builds were approved. Its generated workspace-file edit was reverted; no dependency/config changes remain.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for permission checks across supported roles and actions.
  * Added tests for feature-management permissions, including users, integrations, services, and tags.
  * Verified read-only mode allows viewing while blocking management actions.

### DIFF
--- a/apps/client/src/test/permissions.test.ts
+++ b/apps/client/src/test/permissions.test.ts
@@ -1,0 +1,85 @@
+import { Role } from '@OpsiMate/shared';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { getUserRole } from '@/lib/auth';
+import { isPlaygroundMode } from '@/lib/playground';
+import {
+	canCreate,
+	canDelete,
+	canEdit,
+	canManageIntegrations,
+	canManageServices,
+	canManageTags,
+	canManageUsers,
+	canOperate,
+	canView,
+	hasPermission,
+	isReadOnlyMode,
+	type Permission,
+} from '@/lib/permissions';
+
+vi.mock('@/lib/auth', () => ({ getUserRole: vi.fn() }));
+vi.mock('@/lib/playground', () => ({ isPlaygroundMode: vi.fn() }));
+
+const checks: Record<Permission, () => boolean> = {
+	create: canCreate,
+	edit: canEdit,
+	delete: canDelete,
+	view: canView,
+	operate: canOperate,
+};
+const permissions = Object.keys(checks) as Permission[];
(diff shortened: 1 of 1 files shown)


idanlodzki:
This is the matrix as data — `describe.each(roles)` × `test.each(permissions)` — which is exactly the shape #927 hoped for, and the 62 generated cases cover every role including `null` and `Operation`.

You also caught something **my issue got wrong**: I wrote that playground "grants admin-level permissions." It doesn't — `hasPermission` returns false for everything except `view` when `isReadOnlyMode()`, regardless of role. Your test encodes the real behavior (read-only, blocks even admins), which is the correct thing to pin. I've noted the correction on #927.

Mutation-checked: letting Editors delete fails exactly the Editor/delete case. Prettier and lint clean at `--max-warnings=0`. The `import { Role } from '@OpsiMate/shared'` means this lands cleanly whether it goes in before or after #933. Merging when green. That's four solid PRs from you in two days — thank you.

PR #935, opened 2026-09-11
[FIX]: Test integration avatar source fallbacks and labels

### LINKED ISSUE #925: Add unit tests for IntegrationAvatar.utils (source resolution fallback chain)
## Problem
`apps/client/src/components/Alerts/IntegrationAvatar.utils.tsx` decides which integration icon an alert row shows, via a fallback chain, and has no tests. Note the file is `.tsx`.

## What to test
New file `apps/client/src/test/IntegrationAvatar.utils.test.tsx`:
- the fallback order: explicit `type` → a primary tag → the alert id prefix → the summary → `'custom'`
- matching is substring-based and case-insensitive (e.g. an id like `UPTIMEKUMA_9` resolves to Uptime Kuma; `Grafana` and `grafana` both match)
- every known integration kind returns a label

*(easy)*

sridharkalaibala:
Add tests for the integration-avatar utility's public exports: source fallback order, case-insensitive substring matching and aliases, and labels for all six integration kinds.

Fixes #925.

The resolver now lives in `@OpsiMate/shared` and is re-exported by the TSX utility. The suite imports through that utility, exercising the shared implementation without mocking it and verifying its labels agree with the avatar definitions.

Coverage includes conflicting type/tag/id/summary hints, explicit Custom stopping fallback, missing/unrecognized legacy types, `UPTIMEKUMA_9`, Google and uptime-kuma aliases, unrecognized input, primary-tag selection, and hidden/empty tags.

Validation:
- 28 focused tests passed.
- Mutation checks against the real shared source: removing lowercasing caused 15 failures; ignoring explicit type caused 2 failures. Restored the source byte-for-byte afterward.
- Changed-file ESLint passed with `--max-warnings=0`. The test groups were split to comply with the existing function-length limit.
- No production code, dependency, or configuration changes.

AI assistance: OpenAI Codex helped implement and validate these tests.

Full client suite: 344 tests in 39 files passed on this branch, based on current main independently of the permission-test PR. Prettier and `git diff --check` passed. Used the installed test/lint binaries directly on Windows / Node 24.

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for avatar normalization and alert integration resolution.
  * Verified fallback behavior, visible tag handling, legacy and unrecognized values, and label consistency.

### DIFF
--- a/apps/client/src/test/IntegrationAvatar.utils.test.tsx
+++ b/apps/client/src/test/IntegrationAvatar.utils.test.tsx
@@ -1,0 +1,122 @@
+import { Alert, AlertIntegrationKind, AlertSeverity, AlertStatus } from '@OpsiMate/shared';
+import { describe, expect, test } from 'vitest';
+import {
+	getIntegrationLabel,
+	integrationDefinitions,
+	normalizeIntegration,
+	resolveAlertIntegration,
+} from '@/components/Alerts/IntegrationAvatar.utils';
+
+// Exercise legacy/unrecognized type values that make the fallback chain necessary.
+function alert(overrides: Partial<Omit<Alert, 'type'>> & { type?: string | null } = {}): Alert {
+	return {
+		id: 'unknown-1',
+		type: undefined,
+		status: AlertStatus.FIRING,
+		severity: AlertSeverity.WARNING,
+		tags: {},
+		startsAt: '2026-01-01T00:00:00Z',
+		updatedAt: '2026-01-01T00:00:00Z',
+		createdAt: '2026-01-01T00:00:00Z',
+		alertUrl: '',
+		alertName: 'Example alert',
+		isSilenced: false,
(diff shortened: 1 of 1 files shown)


idanlodzki:
Verified — the precedence chain is tested the way it should be, one step at a time with each source deliberately overridden by the one above it (type → primary tag → id → summary → custom), plus the hidden-tag subtlety I hadn't listed in #925. 28 tests. Mutation-checked: removing the `summary` step from `resolveAlertIntegration` fails exactly the summary case. Prettier and lint clean. Merging when green.

sridharkalaibala:
Thanks for reviewing and mutation-checking this and #934, and for merging them. I appreciate the care you put into verifying that the tests catch actual behavior changes. I'll keep checking the remaining contributions and respond to questions or requested changes here.

PR #936, opened 2026-09-12
test/use-filter-panel-hook-926

Abhishek84313:
## Issue Reference

Closes #926

---

## What Was Changed

Added `apps/client/src/test/useFilterPanel.test.tsx` with unit tests for the `useFilterPanel` hook, covering:

- Initial display limit of 6 facet values per field, with `hasMore`/`remaining` computed correctly (including when facet count is below the limit)
- `handleLoadMore` revealing 5 more values at a time, up to full reveal
- Per-field search being case-insensitive and matching on both `value` and `displayValue`
- Fields being independent — loading more or searching in one field doesn't affect another
- `shouldShowSearch` only returning true once facets exceed the initial display limit

---

## Why Was It Changed

`useFilterPanel` drives the filter sidebar's "show more" and search behavior but had no test coverage, making regressions in pagination or search matching easy to miss.

---

## Screenshots

N/A — hook-level unit tests, no UI changes.

---

## Additional Context (Optional)

- `remaining` is not clamped to zero in the hook's current implementation — when the facet count is below the display limit it can go negative (e.g. `4 - 6 = -2`). Tests assert this actual behavior rather than assuming clamping, since fixing it was out of scope for this issue.
- Verified with `npx vitest run` — full client suite (39 files, 325 tests) passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alert toolbar responsiveness so controls wrap cleanly in narrow layouts instead of being clipped.
  * Prevented alert search fields from overflowing or becoming unusable in cramped toolbars.

* **Tests**
  * Added coverage for filter panel behavior, including facet display limits, load-more functionality, case-insensitive search, independent field state, and search visibility conditions.

* **Documentation**
  * Added documentation describing filter panel test coverage and validation results.

Issue #926, opened 2026-09-11
Add unit tests for the useFilterPanel hook


idanlodzki:
## Problem
`apps/client/src/components/shared/hooks/useFilterPanel.ts` (not the separate `useFilterPanelCollapsed`) drives the filter sidebar's "show more" and search behavior, untested.

## What to test
New file `apps/client/src/test/useFilterPanel.test.tsx` using `renderHook` from `@testing-library/react`:
- initially shows at most 6 facet values per field, with `hasMore` / `remaining` reported correctly
- `handleLoadMore` reveals 5 more at a time
- the per-field search is case-insensitive and matches on both `value` and `displayValue`
- fields are independent — loading more or searching in one doesn't affect another

*(easy)*

PR #937, opened 2026-09-12
Fix/alerts search overlap 920

### LINKED ISSUE #920: Search banner text overlaps the alerts toolbar
## Problem
When you type into the alerts search box, the "Showing only alerts matching …" banner (`apps/client/src/components/Alerts/Alerts.tsx` ~line 1328) renders correctly *below* the toolbar — but the search input's own text also paints *over* the "Silenced" / "Split by owner" toolbar buttons, producing garbled overlapping text at the top of the table.

Reproduce: open the alerts page (playground works: `http://localhost:8080/alerts?playground`), type any alert name into "Search alerts…", look at the toolbar row. The screenshot in PR #915's `alert-rca.png` shows it (the mangled text over "Split by owner").

## Suggested fix
Likely the search input is absolutely positioned or lacks a background/z-index relative to the toolbar, so its text bleeds through when the field is wider than its container. Find the search input in `Alerts.tsx` / the toolbar component and either constrain its width, give it an opaque background, or fix the stacking order. Verify at 1440px and at a narrow viewport.

*(easy — CSS/layout; no logic change)*

Abhishek84313:
## Issue Reference

Closes #920

---

## What Was Changed

Two files, both in the alerts toolbar:

**`apps/client/src/components/Alerts/AlertsTable/SearchBar/SearchBar.tsx`** — keeps the search field's contents inside the field:

- `min-w-0` on the `<input>`, so it can actually shrink with its container
- `min-w-0` + `overflow-hidden` on the bordered wrapper, so nothing inside it (input text, clear button, AI toggle) can paint outside the field
- `min-w-0` on the component root

**`apps/client/src/components/Alerts/Alerts.tsx`** — gives the toolbar somewhere to go when it runs out of room:

- The toolbar row is now `flex-wrap` (with `gap-2` instead of `gap-4`), so controls drop to a second line instead of being clipped
- The search wrapper keeps a `min-w-[192px]` floor instead of `min-w-0`, so the field stays usable rather than collapsing

---

## Why Was It Changed

The `<input>` is a flex child, so it carries the default `min-width: auto` — its automatic minimum size, which for a text field is its intrinsic ~20-character width. Its wrapper, meanwhile, was `flex-1 min-w-0` and free to collapse to nothing.

Every other control in that row is `shrink-0`, so when space runs short the row squeezes the only thing it can: the search wrapper. Measured on the playground at 1440px with the details pane open, the wrapper collapsed to **2px** while the input held its **175px** intrinsic width — **194px of text overflowing**, rendered straight over the "Silenced" and "Split by owner" buttons. That is the garbled toolbar in the issue.

Worth flagging for reviewers: the issue guessed at absolute positioning or a z-index problem. That turned out not to be the cause — nothing in the search field is absolutely positioned and no stacking context is involved. It is purely flexbox minimum-size overflow, which is why the fix is `min-w-0` rather than a `z-index` or `background` change.

The wrapping half addresses the same cramped row from the other side: at 1440px with the details pane open the toolbar's controls simply do not fit on one line, which also pushed the time filter out of view (the clipped "All t…" visible in the before shot).

---

## Screenshots

Alerts toolbar before and after the fix

Captured from the playground (`/alerts?playground`) with the same search term in all three, driven through the Chrome DevTools Protocol so the measurements below come from the live DOM rather than from eyeballing the image.

---

- **Measured result.** Overflow of the input past its box, and any toolbar buttons it lands on:

    | View | Before | After |
    | --- | --- | --- |
    | 1440px, details pane open | +194px over `Silenced`, `Split by owner` | −23px, none |
    | 1024px, details pane open | — | −23px, none |
    | 1440px, no details pane | — | −23px, none |

- **Verification.** `tsc --noEmit` clean; `eslint` reports 0 errors on both files (21 warnings, all pre-existing `react-hooks` complaints about `setState` in effects that predate this change); Prettier clean; full client suite passes — `vitest run`, 39 files, 325 tests.
- **Second consumer.** `SearchBar` is also rendered by the `AlertsTable` toolbar, whose wrapper is a bare `flex-1`. The `min-w-0` changes protect that usage from the same overflow with no change at the call site.
- **`gap-4` → `gap-2`.** Deliberate: the tighter gap buys back horizontal room and keeps the two rows from looking disconnected once the toolbar wraps.
- **`min-w-[192px]`** rather than `12rem` — same value, but every other arbitrary width in the client is expressed in px (`min-w-[350px]`, `min-w-[150px]`, …).
- **The AI-mode hint** (`absolute -bottom-4`) sits on the component root, outside the new `overflow-hidden` wrapper, so it is unaffected.


idanlodzki:
The fix here is exactly right, and the analysis behind it is the best I've seen on this issue: you measured the wrapper collapsing to 2px while the input held its 175px intrinsic width — the classic flex-child `min-width: auto` trap — and `min-w-0` + `overflow-hidden` on the field with `flex-wrap` on the toolbar is the correct cure. The inline comments explaining *why* are exactly what the codebase wants. I want to merge this.

But the PR as it stands can't be merged, for reasons that are about git, not your code:

1. **It's stacked on #936.** This branch contains #936's commits (`3162d60a`, `165f4ffe`, `431101d5`), so it also adds `useFilterPanel.test.tsx` — a different issue's change. Each PR needs to be its own branch off `main`.
2. **`PR_DESCRIPTION.md` and two PNGs are committed to the repo root.** Screenshots belong in the PR description (drag-and-drop into the comment box — GitHub hosts them), never in the tree; a 430KB PNG in the root would ship in every clone forever. Same for the description file.
3. Commit messages like "SS as you asked" won't mean anything in `git log` six months from now.

Cleanest path — start a fresh branch and cherry-pick just the fix:
```
git fetch origin && git checkout -b fix/search-overlap-920 origin/main
git cherry-pick f842390e          # your fix commit only
git push -u origin fix/search-overlap-920
```
then open a new PR from it (or force-push this branch to it), paste the screenshot into the description, and close this one. The 10-line diff to `SearchBar.tsx` + `Alerts.tsx` is all it needs. Ping me and I'll approve CI and merge promptly — this is good work that deserves a clean landing.

idanlodzki:
Checking in — the fix in here is good and I'd like to land it. The cherry-pick recipe in my earlier comment is three commands; if git is being difficult, say so and I'll cut the clean branch myself with you as author. Either way, this and #936 need a fresh branch each — the stacked one with the screenshots committed can't merge. No rush, but I'll close these in a week if there's no movement, and the issues (#920, #926) go back to the pool.

idanlodzki:
Closing this for now to keep the queue honest — it's been quiet since Sep 12 and CI is red on the branch history rather than on the change itself. **This is not a rejection**: the overlap fix is right and I'd merge it the same day from a clean branch (cherry-pick the one commit onto current main, as in my earlier comment). Reopen or open a fresh PR whenever you have time — #920 stays open for you or anyone else. Thanks for the work you put in.

idanlodzki:
Checking in — this is still open for you. The one thing standing in the way is the branch history (CI fails on unrelated commits, not on your change). Cherry-picking your commit onto a fresh branch from current `main` fixes it; if git is being difficult, tell me and I'll cut the clean branch myself with you as the author.

PR #940, opened 2026-09-12
[FIX]: Consolidate tag contrast helper and replace deprecated substr

### LINKED ISSUE #912: Duplicated getContrastColor helper using deprecated String.prototype.substr
## Problem
The same hex-luminance helper is copy-pasted in two places, and both use the deprecated `.substr`:

- `apps/client/src/components/ui/tag-badge.tsx:39-47` (three `.substr` calls at lines 44–46)
- `apps/client/src/components/Dashboards/Dashboards.utils.ts:74-80` (three at 76–78)

These are the only `.substr` uses left in the client.

## Suggested fix
Keep one implementation (the `Dashboards.utils.ts` one is already in a utils module — or move it somewhere shared like `lib/`), switch `.substr(start, len)` to `.slice(start, start + len)`, and import it from both call sites. Add a tiny unit test: `#ffffff` → dark text, `#000000` → light text (pattern: `apps/client/src/test/distributeWidths.test.ts`).

*(trivial–easy)*

sridharkalaibala:
TagBadge and the dashboard utilities contained the same colour-contrast helper, including the client's remaining deprecated `substr` calls. Move the calculation into `lib/colors.ts`, import it in TagBadge, and re-export it from the existing dashboard utility path to preserve that API. Use `slice` for the RGB pairs; the brightness formula and colour choices stay unchanged.

Fixes #912.

Validation: 10 focused tests cover white/black, each primary colour, greys immediately around the brightness threshold, hashless hex values and the existing dashboard export. All 422 client tests across 42 files pass. Compared the old and new helper over 4,096 RGB samples: identical results. Changed-file ESLint passes with zero warnings; Prettier and diff checks pass. No `substr` calls remain in client source.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text color selection for dashboard and tag badge backgrounds, enhancing readability across a wider range of colors.

* **Tests**
  * Added coverage for contrast color handling across common and hexadecimal color values, including colors without a leading `#`.

### DIFF
--- a/apps/client/src/components/Dashboards/Dashboards.utils.ts
+++ b/apps/client/src/components/Dashboards/Dashboards.utils.ts
@@ -71,11 +71,4 @@ export const formatDate = (dateString?: string): string => {
 	}
 };
 
-export const getContrastColor = (hexColor: string): string => {
-	const hex = hexColor.replace('#', '');
-	const r = parseInt(hex.substr(0, 2), 16);
-	const g = parseInt(hex.substr(2, 2), 16);
-	const b = parseInt(hex.substr(4, 2), 16);
-	const luminance = (0.299 * r + 0.587 * g + 0.114 * b) / 255;
-	return luminance > 0.5 ? '#000000' : '#ffffff';
-};
+export { getContrastColor } from '@/lib/colors';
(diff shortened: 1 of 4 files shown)


idanlodzki:
`substr(2,2)`→`slice(2,4)` and `substr(4,2)`→`slice(4,6)` are the correct conversions, one implementation in `lib/colors.ts`, and keeping `Dashboards.utils` as a re-export means nothing else breaks. Mutation-checked: shifting the luminance threshold fails exactly the boundary cases. #912 done. Merging.

PR #939, opened 2026-09-12
[FIX]: Cover dashboard favourite persistence, sorting and search

sridharkalaibala:
Dashboard favourites are persisted in localStorage, but the storage, toggle, sort and search helpers had no regression coverage. Add 19 tests using the actual utilities and jsdom storage:

- Missing/malformed/unavailable storage and JSON round trips, including unrelated storage preservation.
- Add/remove favourites, preserve other IDs, persist the empty list, and recover after corrupt JSON.
- Favourite-first sorting, alphabetic order within both groups, and input non-mutation.
- Case-insensitive partial search across name, description and tags, plus blank/no-match queries and absent optional fields.

Fixes #923. No production or dependency changes; `getContrastColor` is excluded as requested in the issue.

Validation: all 19 focused tests pass. Mutation checks against the real utility: omitting persistence fails 4 tests; sorting the input array directly fails 1. Production source restored afterward. Changed-file ESLint passes with zero warnings and Prettier formatting applied.

Developed and validated with OpenAI Codex assistance.

Full client suite: 431 tests across 42 files pass.

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for dashboard favorites, including saving, removing, and recovering favorite selections.
  - Added validation for favorite-first sorting and alphabetical ordering.
  - Added coverage for case-insensitive dashboard search across names, descriptions, and tags.
  - Verified graceful handling of blank searches, incomplete dashboard details, malformed saved data, and unavailable storage.

PR #942, opened 2026-09-12
[FIX]: Share clipboard handling while preserving dialog and table feedback

### LINKED ISSUE #918: Extract the copy-pasted clipboard handler into a useCopyToClipboard hook
## Problem
Seven components carry a near-identical block — `navigator.clipboard.writeText(...)`, a success toast, and a `setTimeout` that resets a "copied" flag after 2000ms:

- `components/Integrations/GCPSetupModal/GCPSetupModal.tsx`
- `components/Integrations/UptimeKumaSetupModal/UptimeKumaSetupModal.tsx` (byte-for-byte identical to the GCP one)
- `components/Integrations/GrafanaSetupModal/GrafanaSetupModal.tsx`
- `components/Integrations/DatadogSetupModal/DatadogSetupModal.tsx`
- `components/Integrations/CustomAlertsSetupModal/CustomAlertsSetupModal.tsx`
- `components/Integrations/ZabbixSetupModal/ZabbixSetupModal.tsx`
- `components/Alerts/AlertsTable/AlertRow/CopyCellButton.tsx`

(all under `apps/client/src/`). Any change — a different toast, a longer reset — means seven edits.

## Suggested fix
Add `apps/client/src/hooks/useCopyToClipboard.ts` exporting a hook that returns `{ copied, copy }`, with the reset delay as a named constant (`COPIED_RESET_MS = 2000`). Replace all seven blocks with it. Keep behavior identical — same toast text per call site. A small test (mock `navigator.clipboard.writeText`, assert `copied` flips true then false after the delay with fake timers) would be a good addition.

*(easy)*

sridharkalaibala:
Seven components independently implement copying and temporary success feedback. Introduce `useCopyToClipboard` with a named `COPIED_RESET_MS` default and move all seven call sites onto it. The hook owns clipboard access, optional fallback, toast feedback and timer cleanup. Multi-button dialogs supply callbacks to retain their existing indicator grouping.

Fixes #918.

The current implementations differ from the issue's description, so this preserves their observable distinctions:
- Integration dialogs retain their exact success/error toast messages, 2-second success/reset duration and 3-second error duration.
- Datadog's copy deadline still resets both indicators together; CustomAlerts and Zabbix retain their independent indicator callbacks.
- CopyCellButton retains its 1.5-second feedback, restart-on-repeat behaviour, no toast, click/mousedown propagation guards and legacy execCommand fallback for unavailable/rejected Clipboard API.

Timers are now cleaned up for dialogs too, and async copies finishing after unmount do not trigger feedback. A throwing legacy fallback is treated as failure instead of an unhandled rejection.

Validation:
- 24 tests cover the real table button, all six integration dialogs and shared hook: copied values, success/error text, reset boundaries/grouping, fallback failure/cleanup, repeated copies, unmount and pending promises. All 18 component regression tests also pass against the original upstream components.
- Full client suite: 436 tests across 45 files pass.
- TypeScript baseline comparison: no new diagnostics; the old CopyCellButton useRef-without-an-argument error is removed (25 baseline diagnostics to 24). The overall project type check remains affected by unrelated baseline errors.
- ESLint: no errors; touched-file warnings decrease from 14 to 9 existing size-limit warnings. New hook/tests have zero warnings. Prettier and diff checks pass.
- Browser clipboard permissions and real integration services were not exercised; clipboard operations are mocked in jsdom.

### DIFF
--- a/apps/client/src/components/Alerts/AlertsTable/AlertRow/CopyCellButton.tsx
+++ b/apps/client/src/components/Alerts/AlertsTable/AlertRow/CopyCellButton.tsx
@@ -1,30 +1,6 @@
 import { Check, Copy } from 'lucide-react';
-import { useEffect, useRef, useState } from 'react';
+import { useCopyToClipboard } from '@/hooks/useCopyToClipboard';
 import { cn } from '@/lib/utils';
-
-// The Clipboard API needs a secure context, and self-hosted OpsiMate commonly runs on
-// plain http — fall back to the legacy execCommand path there.
-const copyText = async (text: string): Promise<boolean> => {
-	if (navigator.clipboard?.writeText) {
-		try {
-			await navigator.clipboard.writeText(text);
-			return true;
-		} catch {
-			// Permission denied or transient failure — try the fallback below.
-		}
-	}
-	const textarea = document.createElement('textarea');
-	textarea.value = text;
-	textarea.style.position = 'fixed';
-	textarea.style.opacity = '0';
-	document.body.appendChild(textarea);
-	textarea.select();
-	try {
-		return document.execCommand('copy');
-	} finally {
-		textarea.remove();
-	}
-};
 
 export interface CopyCellButtonProps {
 	// The FULL underlying value — not the (possibly truncated/reformatted) display text.
@@ -36,16 +12,11 @@ export interface CopyCellButtonProps {
 // `relative group/cell` for the reveal and positioning to work. Click/mousedown stop
 // propagation so copying never opens the details panel or starts a drag selection.
 export const CopyCellButton = ({ value, className }: CopyCellButtonProps) => {
-	const [copied, setCopied] = useState(false);
-	const resetTimer = useRef<number>();
-	useEffect(() => () => window.clearTimeout(resetTimer.current), []);
+	const { copied, copy } = useCopyToClipboard({ resetMs: 1500, restartTimer: true, legacyFallback: true });
 
 	const handleCopy = async (e: React.MouseEvent) => {
 		e.stopPropagation();
-		if (!(await copyText(value))) return;
-		setCopied(true);
-		window.clearTimeout(resetTimer.current);
-		resetTimer.current = window.setTimeout(() => setCopied(false), 1500);
+		await copy(value);
 	};
 
 	return (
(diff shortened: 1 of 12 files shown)


idanlodzki:
Verified: all seven copy-pasted blocks are gone (zero `navigator.clipboard.writeText` calls outside the hook), client 436/436, prettier clean, typecheck gate green, and the lint warning count on the touched files went **14 → 0** — the refactor cleaned up as it went.

The `http://` fallback is a genuinely good catch that #918 didn't ask for: the Clipboard API requires a secure context, and self-hosted OpsiMate on plain http is common, so the `execCommand` path means copy actually works for those users. Thank you for that.

One honest design note, not a blocker: the hook grew three knobs (`restartTimer`, `legacyFallback`, per-call callbacks) to preserve each call site's exact prior behavior. That's the *safe* way to dedupe — nothing changes for users — but it also means the seven copies had drifted and the drift is now encoded as configuration. A follow-up that picks one behavior (I'd argue: always restart the timer, always fall back) and deletes the knobs would make this hook half the size. Not for this PR; it's correct as is. Merging.

sridharkalaibala:
Thanks for the detailed review and the validation across this batch, including the mutation checks on #939 and #940.

Agreed on making the clipboard behavior consistent in a separate follow-up. I will leave this PR as reviewed, then work from the merged version on always restarting copied feedback and always attempting the fallback. One detail to preserve in that follow-up: the fallback is currently opted into by the table copy button; enabling it for the dialogs is an intentional behavior improvement. The multi-button dialogs also need their indicators handled explicitly when the per-call callbacks are removed, so I will cover repeated copies and switching between buttons as well as fallback failure and unmount cleanup.

idanlodzki:
@sridharkalaibala — stepping outside this PR for a second.

Thirteen PRs in three days, every one of them clean on the first or second pass: the tests actually bite under mutation, the refactors leave zero stragglers, and #931 was a real hole in our CI gate that nobody had asked anyone to find. That's not "good first issue" work anymore.

**Idan (the maintainer) would like to talk with you on Slack about taking on something bigger** — there's real feature work on the roadmap (notification routing, the AI investigation agent, incidents) that needs someone who works the way you do. Slack invite is in the README badge: https://join.slack.com/t/opsimate/shared_invite/zt-39bq3x6et-NrVCZzH7xuBGIXmOjJM7gA — say hi in the channel and mention this comment.

Thank you for the work so far. Genuinely.

sridharkalaibala:
Thank you for the invitation and for the careful review. I've passed the Slack invitation and your roadmap suggestions to the account owner so they can follow up personally.

For clarity, these contributions are prepared with AI assistance, as disclosed in the PRs. I can continue the technical investigation and review follow-ups here on GitHub; the personal Slack conversation is for them. The separate clipboard simplification is prepared locally and will be based on the merged #942 before submission.

sridharkalaibala:
The separate follow-up is ready in #956, based on current main after this merge. It always restarts copied feedback, enables fallback for all consumers, and removes the behavior knobs/callbacks with independent indicators for multi-button dialogs. All 476 client tests and the production build pass; the type baseline reports no new errors. The PR describes the intentional dialog fallback and Datadog timer behavior changes. Thanks again for the review direction.

PR #943, opened 2026-09-12
[FIX]: Attribute secrets endpoint logs to the correct controller

### LINKED ISSUE #907: Secrets controller logs under the wrong module name
## Problem
`apps/server/src/api/v1/secrets/controller.ts:10`:

```ts
const logger = new Logger('v1/integrations/controller');
```

The secrets controller was copy-pasted from the integrations one and kept its logger context, so every log line from the secrets endpoints is attributed to integrations. Anyone grepping logs for a secrets problem looks in the wrong place.

## Suggested fix
Rename the context to match the module-path convention the other controllers use, e.g. `'api/v1/secrets/controller'` (see `apps/server/src/api/v1/audit/controller.ts` for the pattern).

*(trivial — one-line change; a good first PR to learn the review flow)*

sridharkalaibala:
Errors from the secrets endpoints are currently tagged `v1/integrations/controller`, directing log searches to the wrong module. Change the logger context to `api/v1/secrets/controller`, matching the module-path convention used by the audit controller.

Fixes #907.

Validation: verified the logger string is the only source change; TypeScript syntax/transpilation, Prettier and diff checks pass. Server ESLint could not run because this local server workspace lacks `@eslint/js`. No endpoint behaviour changes; full server tests were not run for this logging-label-only fix.

### DIFF
--- a/apps/server/src/api/v1/secrets/controller.ts
+++ b/apps/server/src/api/v1/secrets/controller.ts
@@ -7,7 +7,7 @@ import { encryptPassword } from '../../../utils/encryption';
 import { isZodError } from '../../../utils/isZodError';
 import { validateKubeConfig, validatePublicSSHKey } from '../../../utils/validators/validators';
 
-const logger = new Logger('v1/integrations/controller');
+const logger = new Logger('api/v1/secrets/controller');
 
 export class SecretsController {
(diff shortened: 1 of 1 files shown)


idanlodzki:
One line, exactly as #907 asked. Verified. Merging.

Issue #896, opened 2026-08-31
Flaky server suite: full parallel runs fail ~1-in-4 under machine load, rotating files


idanlodzki:
## Problem
Full `vitest run` of the server suite fails roughly 1 run in 4 when the machine is otherwise busy, and passes when rerun. The failing test **rotates between files** — each observed failure passes in isolation immediately afterwards:

- `dashboards.test.ts > both toggles survive a create as real booleans` (2026-08-31, on the #893 branch — before any of the perf changes)
- `alerts.test.ts > a 'contains' matcher matches by case-insensitive substring` and `auth.test.ts > should allow access to protected endpoint with valid JWT` (2026-08-31, on #892's branch, while the client suite ran concurrently)
- `auth.test.ts > should update user profile successfully` and once `queryResultCache.test.ts` (2026-09-01, on the #895 branch)

Observed across three different branches, so it is not introduced by any one of them.

## What's known
- `vitest.config.ts` already pins `pool: 'forks'` for process isolation, and `setupDB()` gives each file its own `:memory:` database — so it is not shared-DB or shared-env contamination in the obvious sense.
- Failures correlate with overall machine load (concurrent client suite, dev servers, load tests), which points at timing sensitivity rather than logic: bcrypt hashing in register/login paths (~100ms CPU each) inside parallel forks is a suspect, as are worker-reuse effects on module-level singletons.
- The per-test timeout is 30s, so simple slowness shouldn't fail assertions — the failures are assertion failures, not timeouts, which is the puzzling part. Nobody has yet captured a full assertion diff from a flaked run.

## Suggested next step
Capture the actual assertion output on a flaked run (run the suite in a loop with `--reporter=verbose` teeing to a file until it fails), before attempting any fix. If the flakes turn out to be worker-reuse of module singletons, `poolOptions.forks.isolate: true` (or `singleFork` for the few app-constructing files) is the likely knob.

*(fills the gap noted in passing on #893 and #895 — filing it so it stops being folklore)*

idanlodzki:
Rate update: today's verification runs on main (post #903/#904 merges) failed 3 of 4 full runs, each a different test (`integrations > should require authentication`, `alerts > unresolve timeline`, `alerts > unsilence 401`), all passing in isolation. Same signature as before, but the frequency is now the majority of runs on this machine under load — it's past 'occasional'. Suite is at 24 files / 356 tests, so it has grown ~15% since the issue was filed, which fits a worker-contention theory.

[...]

PR #944, opened 2026-09-12
[FIX]: Safely parse Datadog webhook timestamps

### LINKED ISSUE #766: Datadog webhook can return 500 on a non-numeric timestamp
## Problem
In `apps/server/src/api/v1/alerts/controller.ts` (lines ~387-388), the Datadog webhook does `new Date(Number(startsAtSource)).toISOString()`. The zod schema only requires `z.string()` for the date field, so if Datadog sends an ISO/text date, `Number(...)` yields `NaN`, the `Date` is invalid, and `.toISOString()` throws a `RangeError` — turning a recoverable payload into a 500.

## Suggested fix
The same file already has a safe helper — `AlertController.toIsoOrNow` (line ~430). Reuse it for both `startsAt` and `updatedAt`.

## Acceptance criteria
- [ ] A Datadog payload with an ISO-string date is accepted (no 500)
- [ ] Numeric epoch timestamps keep working

Swayamcodes:
## Issue Reference

Closes #766

---

## What Was Changed

* Added Datadog-specific timestamp normalization that accepts ISO dates while preserving existing numeric-string epoch conversions, including decimals.
* Applied normalization to both `startsAt` and `updatedAt`.
* Added endpoint regression coverage for ISO, numeric, decimal, missing `last_updated`, and invalid timestamps.

---

## Why Was It Changed

ISO timestamps previously passed payload validation but caused `Number(value)` to produce `NaN`, making `toISOString()` throw and returning HTTP 500.

The fix reuses the existing safe date parser for non-numeric values without changing Grafana behavior or Datadog’s timestamp-field precedence. Invalid timestamps now safely fall back to the current time.

---

## Screenshots

Not applicable — backend-only change with no visual UI changes.

---

Validation on the rebased branch:

* Full server suite: 370 tests passed across 25 files
* Full alerts test file: 101 tests passed
* Server build passed
* Focused Prettier check passed
* Git whitespace checks passed

Controller ESLint passed. The alerts test file retains its existing 183 errors and one warning; no additional diagnostics were introduced.

## Summary by CodeRabbit

- **Bug Fixes**
  - Datadog alert ingestion now handles epoch timestamps, including decimal values, as well as ISO-formatted timestamps.
  - Missing update timestamps fall back to the alert’s creation timestamp.
  - Invalid, blank, or whitespace-only timestamps no longer cause ingestion failures and instead use the current time.
  - Zero-valued timestamps are preserved correctly as the Unix epoch.
  - These improvements provide more reliable alert timestamps across supported input formats and prevent malformed timestamp data from interrupting ingestion.

### DIFF
--- a/apps/server/src/api/v1/alerts/controller.ts
+++ b/apps/server/src/api/v1/alerts/controller.ts
@@ -632,6 +632,8 @@ export class AlertController {
 	// ISO timestamps too. Preserve Number()'s accepted numeric forms (including decimals),
 	// then delegate ISO and invalid values to Grafana's established fallback parser.
 	private static toDatadogIsoOrNow(value?: string): string {
+		if (value?.trim() === '') return AlertController.toIsoOrNow(undefined);
+
 		const epochMs = Number(value);
 		if (value !== undefined && Number.isFinite(epochMs)) {
 			const parsed = new Date(epochMs);
--- a/apps/server/src/api/v1/alerts/controller.ts
+++ b/apps/server/src/api/v1/alerts/controller.ts
@@ -561,8 +561,8 @@ export class AlertController {
 				// (P1 → critical, …) unless an explicit severity tag is present.
 				severity: tags['severity'] ?? payload.priority,
 				tags,
-				startsAt: new Date(Number(startsAtSource)).toISOString(),
-				updatedAt: new Date(Number(updatedAtSource)).toISOString(),
+				startsAt: AlertController.toDatadogIsoOrNow(startsAtSource),
+				updatedAt: AlertController.toDatadogIsoOrNow(updatedAtSource),
 				alertUrl: payload.link ?? '',
 				alertName: payload.title || 'UNKNOWN',
 				summary: payload.message,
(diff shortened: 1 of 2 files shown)


idanlodzki:
Good fix, and more careful than the issue's own suggestion. #766 said "just reuse `toIsoOrNow`" — which would have quietly broken Datadog's **default** format (epoch milliseconds as a string), since `toIsoOrNow` would parse `"1757500000000"` as an ISO string and fall back to now. Your `toDatadogIsoOrNow` tries numeric first and only then delegates, so epoch keeps working and ISO/garbage no longer 500s. That's the right shape, and the test matrix (epoch, ISO, padded/decimal zero, invalid → now) pins it.

Verified: server suite green, lint at `--max-warnings=0` clean, prettier clean. One suggestion, not blocking: the `padded zero` and `decimal zero` cases both normalize to the same alert id (`datadog-0-timestamp`), so under parallel test load they can race on one row — I saw one such flake in 4 full runs (passes in isolation and on rerun; it's our known #896 flakiness, but a distinct id per case, e.g. including `$name`, would make yours immune). Happy to merge either way; ping me if you push that tweak and I'll re-approve the gate. Third PR from you — thanks, and if you haven't, a ⭐ helps.

idanlodzki:
Your `934e604` push landed the blank-timestamp guard — good, that was a real one (a whitespace-only date would have persisted as 1970-01-01).

One thing still open from CodeRabbit, and it's a repo rule rather than a nit: **no anonymous inline object types, tests included** — the six `as { starts_at: string; updated_at: string }` / `as { id: string; … }` casts in `alerts.test.ts` (lines ~1043–1143). It's enforced from the org's CodeRabbit config, so it isn't discoverable from the repo, which is why it catches most first-time contributors. Fix is two interfaces at the top of the Datadog `describe` block:

```ts
interface DatadogTimestampRow {
	starts_at: string;
	updated_at: string;
}
interface DatadogAlertRow {
	id: string;
	// …whichever fields that first cast selects
}
```
then `as DatadogTimestampRow` / `as DatadogAlertRow` at each site. While you're in there, giving the `padded zero` / `decimal zero` cases distinct alert ids (e.g. include `$name` in the id) would make the test immune to the parallel-run race I mentioned. Push, ping me, and it merges — the fix itself is done.

Swayamcodes:
@idanlodzki ive pushed the requested updates, the Datadog database-row casts now use named interfaces, and all zero-timestamp parameterized cases use distinct alert IDs. The full server suite passes with 374/374 tests, and the focused Datadog tests pass 13/13. and yes ive already starred.

PR #946, opened 2026-09-12
[FIX]: Name the shared bcrypt password hashing work factor

### LINKED ISSUE #908: bcrypt cost factor is a magic 10 repeated four times
## Problem
`apps/server/src/bl/users/user.bl.ts` calls `bcrypt.hash(password, 10)` at lines 35, 52, 77 and 120 — the cost factor is a bare literal in each. Changing it (e.g. raising it as hardware gets faster) means finding every copy, and nothing documents what the number is. No `SALT_ROUNDS` identifier exists anywhere in `apps/server/src`.

## Suggested fix
Extract a module-level `const BCRYPT_SALT_ROUNDS = 10;` at the top of `user.bl.ts` with a one-line comment on what it is, and use it at all four sites. `apps/server/src/utils/encryption.ts` names its crypto constants the same way — match that style. Making it config-driven is optional and out of scope for a first pass.

*(trivial)*

sridharkalaibala:
Fixes #908.

Name the existing bcrypt work factor as `BCRYPT_SALT_ROUNDS = 10` and use it for registration, user creation, password reset, and profile password changes. Add one explanatory comment. The cost remains 10 in all four flows.

Validation: Prettier, TypeScript transpilation/syntax, and `git diff --check` pass. Compared the source after substituting the constant back to `10` with the original file to confirm there are no other changes. No new tests or full server suite for this constant-only refactor; server ESLint was not run because the existing local server installation lacks `@eslint/js`.

Prepared with OpenAI Codex assistance.

### DIFF
--- a/apps/server/src/bl/users/user.bl.ts
+++ b/apps/server/src/bl/users/user.bl.ts
@@ -6,6 +6,9 @@ import { PasswordResetsRepository } from '../../dal/passwordResetsRepository';
 import { AuditBL } from '../audit/audit.bl';
 import { decryptPassword, generatePasswordResetInfo, hashString } from '../../utils/encryption';
 
+// Work factor used for every password hash created by this module.
+const BCRYPT_SALT_ROUNDS = 10;
+
 const logger = new Logger('bl/users/user.bl');
 
 export class UserBL {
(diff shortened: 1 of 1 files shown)


idanlodzki:
`BCRYPT_SALT_ROUNDS` at all four sites, matching `encryption.ts`'s naming — exactly #908. Server 365/365, lint at `--max-warnings=0` clean. Merging.

PR #948, opened 2026-09-12
[FIX]: Remove orphaned alerts heatmap helpers

### LINKED ISSUE #911: Remove the orphaned AlertsHeatmap folder (constants/types/utils with no component)
## Problem
`apps/client/src/components/Alerts/AlertsHeatmap/` contains only `AlertsHeatmap.constants.ts`, `AlertsHeatmap.types.ts` and `AlertsHeatmap.utils.ts` — no component, no `index.ts`, and nothing outside the folder imports any of it. It's the leftover of a feature that was never finished (or was removed without its helpers).

## Suggested fix
Delete the folder. Before you do, confirm with a grep that nothing references it: `grep -rn "AlertsHeatmap" apps/client/src` should return only files inside the folder itself. If you'd rather *finish* the heatmap than delete it, open a discussion first — that's a different, much larger task.

*(trivial — deletion plus one grep to prove it's safe)*

sridharkalaibala:
## Summary

Removes the three unused `AlertsHeatmap` helper files left behind after the heatmap component was removed. The folder had no component or entry point, and its imports were entirely internal. Updates the code-style guide example to the existing `AlertsTable` component.

Fixes #911.

## Validation

- Searched tracked source and configuration for the folder name and unique exports before/after removal; no remaining references. Analytics uses its own separate `SEVERITY_COLORS` export.
- Client Vitest suite: 412 tests passed across 41 files.
- Prettier check for the edited guide and `git diff --cached --check` passed.
- No new tests added for deleting unreachable code. Full application build was not rerun.

## Summary by CodeRabbit

- **Documentation**
  - Updated the component-structure example in internal development guidance.

- **Refactor**
  - Simplified internal alert visualization implementation details without introducing new end-user functionality.
  - No user-facing behavior changes are included in this update.


idanlodzki:
Verified zero remaining `AlertsHeatmap` references on the branch — and you caught the stale example in the Cursor rules file that pointed at the deleted folder, which I hadn't. #911 done. Merging.

PR #953, opened 2026-09-13
[FIX]: Recognize Windows TypeScript diagnostics in the client gate

sridharkalaibala:
The client typecheck gate cannot recognize CRLF-terminated TypeScript diagnostics on Windows: splitting only on `\n` leaves `\r`, so the anchored diagnostic patterns fail. A baselined run is rejected as a compiler launch failure; mixed LF/CRLF output can also hide a new error. TypeScript's forward-slash repository paths additionally fail to match Windows' native backslash prefix.

Normalize CRLF/LF line boundaries and both repository path forms before comparing diagnostic identities. The compiler invocation, baseline file, and safeguards against failed compiler launches remain unchanged. Follow-up to #931, found while validating the separate clipboard work.

Validation:
- Before: five new CRLF regressions fail while the original nine gate tests pass. The forward-slash path regression separately fails on Windows after only the newline fix.
- After: all 16 gate tests and all 429 client tests pass. Cases include baselined exits 1/2, global diagnostics, mixed line endings, baseline updates, and both path forms. Existing package-manager/empty-output failure tests remain passing.
- Touched-file ESLint, Prettier check, and `git diff --check` pass.
- Real Windows compiler output now passes the unchanged baseline comparison after rebuilding the local shared package: no new client type errors (25 recorded baseline identities, one no longer present). For that local check only, the gate's compiler command was temporarily changed from `pnpm exec tsc` to the installed compiler's `node .../tsc` command because local pnpm attempted an install and stopped on ignored dependency build scripts. The command was restored; no tooling policy or baseline changes are in this PR.

Prepared with AI assistance.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cross-platform type-checking by consistently handling Unix and Windows line endings.
  - Normalized repository paths across slash styles to ensure diagnostics match correctly.
  - Prevented new compiler errors from being incorrectly overlooked when compared with existing baselines.

- **Tests**
  - Added coverage for Windows-formatted compiler output and cross-platform repository paths.


idanlodzki:
Verified, and this is a sharp follow-up to your own #931. The strict check we added there — reject a failed `tsc` with no recognized diagnostics — interacts badly with Windows: `tsc` emits CRLF, `split('\n')` leaves a trailing `\r`, the anchored patterns match nothing, and the gate rejects every Windows run as a compiler launch failure. Nobody on macOS/Linux could have hit it, which is exactly why it needed you.

I couldn't fake CRLF at the process level from here, so I ran **your 16 tests against main's gate**: the 7 CRLF/path cases fail, the 9 originals pass. Against your branch: 16/16. That's the cleanest possible proof the fix is targeted. Real gate still passes on current main (`No new client type errors`), prettier and lint at `--max-warnings=0` clean.

`split(/\r?\n/)` + normalizing both path-separator forms is the minimal correct fix. Merging when CI lands.

sridharkalaibala:
Thank you for checking the regression tests against the original gate and for merging this. The Windows output normalization is now also helping validate the clipboard follow-up in #956: the current gate reports no new client errors with the installed Windows compiler. The baseline remains unchanged.

PR #956, opened 2026-09-13
[FIX]: Make clipboard feedback and fallback consistent

sridharkalaibala:
Follow-up requested in the review of #942: https://github.com/OpsiMate/OpsiMate/pull/942#issuecomment-5652291320.

Make clipboard behavior consistent after the original refactor merged. Each successful copy now restarts its feedback timer, and every consumer attempts the legacy fallback when the Clipboard API is missing or rejects. Remove restartTimer, legacyFallback and per-call reset callbacks. Multi-button dialogs use separate hook instances for independent feedback; Datadog no longer resets both indicators at the first deadline. The table retains its 1500 ms delay, full copied value and propagation handling.

This intentionally extends fallback support from the table to the integration dialogs. Failure messages remain caller-specific, unsuccessful copies do not report success, and temporary textareas/timers are cleaned up. Late copy results after unmount are ignored.

Validation on current main (dea5b6a), including merged #942 and #953:
- All 476 client tests in 47 files pass. Regression coverage includes repeated default copies, missing/rejected Clipboard API, fallback returning false or throwing, independent dialog deadlines and unmount cleanup. During preparation, four new behavior regressions failed against the original hook/consumers.
- Client production build passes; touched-file ESLint passes with 6 existing size warnings; Prettier and git diff --check pass.
- Type baseline passes: no new client errors, 25 known baseline identities with 2 no longer occurring. Used the installed compiler directly in the gate to avoid this environment's pnpm auto-install/build-policy problem; temporarily changed only its launch command and restored the script byte-for-byte. No baseline or invocation changes in this PR.

Prepared with AI assistance. No native/system clipboard or live integration service was used in the mocked clipboard tests.

## Summary by CodeRabbit

- **Improvements**
  - Clipboard copying is now more reliable, including when the modern clipboard feature is unavailable or fails.
  - Copying multiple values in setup dialogs provides independent “Copied” feedback and reset timing for each item.
  - Repeated copies correctly refresh their confirmation state, while copy failures provide appropriate feedback.


idanlodzki:
Reviewed and verified on a fresh checkout: client 476/476, typecheck gate green (2 baseline errors gone), prettier clean, lint on touched files shows only the 6 pre-existing size warnings. Also drove the Custom Webhook dialog in the playground build: two copies within 250 ms give two independent `Copied` states, both reset at 2 s, clipboard holds the second value, no console errors.

This is exactly the follow-up I hoped for — one behavior, knobs deleted, hook about half the size, and the test that would have caught the old drift (`restarts the default feedback deadline after another successful copy`) is now in the suite. The unmount race test is a nice touch.

Two nits, not worth a round-trip — noting them for next time:
- `ZabbixSetupModal`: `const { copied: copied, ... }` is a no-op rename.
- `ReturnType<typeof useCopyToClipboard>['copy']` works, but exporting a named `CopyFn` type from the hook would read better at the call site.

Merging. Thanks, Sridhar.

sridharkalaibala:
Thanks for the fresh-checkout checks and the browser verification, and for merging the follow-up. I’ve noted the redundant `copied` alias and the named `CopyFn` suggestion for the next relevant change to this code. The repeated-copy and unmount regressions are now covered for future changes too.

PR #963, opened 2026-09-13
test: cover alert severity normalization

### LINKED ISSUE #961: Add unit tests for shared normalizeAlertSeverity
## Problem
`packages/shared/src/types.ts` `normalizeAlertSeverity` maps every integration's severity vocabulary (Zabbix "disaster", Datadog "P1", Kuma "high"…) onto OpsiMate's three levels. It is only exercised indirectly by an HTTP-level server test, so a synonym-table regression would surface as a webhook test failure, not a unit failure.

## What to test
`packages/shared` has no vitest setup — put the test in `apps/client/src/test/normalizeAlertSeverity.test.ts` (the `@OpsiMate/shared` alias resolves to source there; don't stand up new tooling):
- synonym table: `crit`/`P1`/`high`/`disaster`/`error` → critical; `warn`/`average`/`moderate` → warning; `ok`/`P5`/… → info (read the table in `SEVERITY_SYNONYMS` for the full list)
- case- and whitespace-insensitive: `"  CRIT "` → critical
- `null`, `undefined`, `''` → the default severity
- unknown strings → default
- prototype keys are not treated as synonyms: `'__proto__'`, `'constructor'`, `'toString'` → default (the guard is `Object.hasOwn`; this test pins it)

*(easy — pure function, one `test.each` table does most of it)*

Voyagerroc-Lab:
Closes #961

Add focused client-side unit coverage for integration severity synonyms, case and whitespace normalization, defaults, and prototype-key safety. The pure shared function is exercised without an HTTP server.

Validation: the repository's Vitest binary is not installed in this environment, so the targeted test command could not run locally. git diff --check passes.

## Summary by CodeRabbit

- **Tests**
  - Added automated coverage for alert severity normalization.
  - Verified support for severity aliases, case-insensitive input, and surrounding whitespace.
  - Confirmed invalid, missing, and prototype-related values safely fall back to warning severity.


idanlodzki:
Verified locally and in CI: all 15 cases pass, lint clean, typecheck gate green; CodeRabbit had no findings. Covers everything #961 asked for — the synonym table, fallback, case/whitespace, and the prototype-key guard. The only thing CI tripped on was formatting, so I pushed a Prettier-only commit on top of your branch rather than bounce it back (`pnpm --filter @OpsiMate/client format-fix` before pushing next time and you'll never hit it). Merging — thanks! If you like the project, a ⭐ on the repo helps us a lot.

PR #966, opened 2026-09-13
test: cover alert history selection

### LINKED ISSUE #958: Add unit tests for alertHistory.utils range filtering and the synthesized-update fallback
## Problem
`apps/client/src/components/Alerts/AlertDetails/AlertHistoryTimeline/alertHistory.utils.ts` is partially tested: `alertHistoryGraph.test.ts` covers the graph/lane/CSV helpers. Two exported functions that decide *which* history rows the timeline shows are untested.

## What to test
New file `apps/client/src/test/alertHistorySelection.test.ts`:
- `filterHistoryByRange`: entries exactly on `from` and `to` are included (inclusive bounds); an entry one ms outside either bound is excluded; a `null`/`undefined` range ("All time") returns everything unchanged
- `selectHistoryEntries`: when any real event (non-`UPDATED`) falls inside the window, only real events are returned; when the window hides every real event, the synthesized last-`UPDATED` entry is returned as a fallback; on "All time" with real events present, the fallback stays out; with ONLY the synthesized `UPDATED` entry (status history purged by retention), All-time still shows it — an empty timeline for a live alert is the bug, not the feature

*(easy — pure functions over `AlertHistoryData[]`; build fixtures with fixed ISO timestamps)*

Voyagerroc-Lab:
Closes #958

Add unit coverage for inclusive history range boundaries, All time passthrough, and the synthesized UPDATED fallback when real events fall outside the active window.

Validation: git diff --check passes. Vitest is not installed in this environment, so the targeted test command could not run locally.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for alert history selection, including inclusive date-range boundaries.
  - Verified behavior when no date range is selected.
  - Confirmed updated entries are used as a fallback only when applicable events fall outside the selected period.


idanlodzki:
@coderabbitai review

idanlodzki:
Thanks — the two behaviors you picked are exactly the ones #958 cared about, and the inclusive-bounds and fallback assertions are right. CI is green now: the only failure was formatting, so I pushed a Prettier-only commit on top of your branch (`pnpm --filter @OpsiMate/client format-fix` before pushing next time and you'll never hit it).

One thing left, from CodeRabbit's inline comment and worth doing — two edge cases the suite can't currently catch a regression in:
- an entry **one millisecond before `from`** (your third fixture is a full day past `to`, so an off-by-one on the lower bound would still pass)
- **All time (`null` range) when `data` has only `UPDATED` entries** — the contract is that the synthesized fallback is never used on All time, even if that leaves nothing; right now the All-time assertion has a real event in the data, so it can't tell.

Two extra `expect`s and this merges. If you like the project, a ⭐ on the repo helps us a lot.

idanlodzki:
Gentle ping — CI is green; two extra `expect`s (an entry 1 ms before `from`, and All-time with only `UPDATED` entries) and this goes to maintainer review.

idanlodzki:
Closing this one: #958 has now been covered and merged via #998, which includes the two edge cases we discussed here (1 ms before `from`, All-time with only an `UPDATED` entry) plus the range-filter cases. Nothing wrong with your work — it went quiet for a week and someone else finished the same issue. #964 and #965 are still yours if you'd like to land them; each needs only the small additions in my earlier comments. Thanks!

PR #967, opened 2026-09-14
[FIX]: Reset AI search time range

### LINKED ISSUE #962: AI search keeps the previous time window when the new query has no time phrase
## Problem
In `apps/client/src/components/Alerts/Alerts.tsx`, `applyAiFilter` replaces the filters and search text with the AI result, but only touches the time range when the result carries `lastMinutes`:

```ts
if (result.lastMinutes !== undefined) {
	updateDashboardField('timeRange', minutesToTimeRange(result.lastMinutes));
}
```

The server only sets `lastMinutes` for time phrases ("last hour", "today"). So: ask "critical alerts in the last hour", then ask "database alerts" — the second result silently inherits the 1-hour window from the first, even though everything else was replaced. The user sees a filtered list with no indication that a window is still applied.

## Fix
When `lastMinutes` is absent, reset the time range to the app's empty/default representation (the same value the "All time" control produces — check how `DashboardContext` initialises `timeRange`) so an AI query always describes the whole view. Add a test in `apps/client/src/test/` that applies two results in sequence and asserts the window is cleared by the second.

*(easy — surfaced by CodeRabbit on #868; the code is on main)*

Rayan-and-beyond:
## What changed

- make AI filter results always replace the current time range
- reset to the dashboard's All Time range when `lastMinutes` is absent
- move the minute-to-preset mapping into a small helper so the behavior can be tested directly
- add a regression test for a timed AI query followed by one with no time phrase

## Testing

- focused Vitest regression: 1 passed
- mutation of the no-time fallback back to `last1h` makes the regression fail
- touched files pass Prettier
- touched files pass ESLint with 0 errors (existing `Alerts.tsx` warnings remain)
- client typecheck reports 0 new errors (25 known baseline errors)
- `git diff --check`

## UI context

The affected control is the Alerts dashboard time-range selector shown at the upper right. An AI search without a time phrase now resets this control back to **All time**.

Alerts dashboard showing the All time control

Fixes #962

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AI-powered alert filtering to apply recognized time ranges consistently.
  * AI filters now clear the previous AI time window when no time range is specified.
  * Time durations are mapped to the closest supported preset, including longer ranges.

* **Tests**
  * Added coverage for AI time-range conversion and clearing behavior.


idanlodzki:
Reviewed and verified locally on top of main: tests, lint, Prettier and the client typecheck gate all clean; CodeRabbit had no findings. Nice touch moving the preset mapping into its own util — that's what made the sequence testable, and the empty range you reset to is exactly what `createEmptyTimeRange()` produces, so the AI path and the All-time control agree. Merging — thanks! If you like the project, a ⭐ on the repo helps us a lot.

PR #968, opened 2026-09-15
[FIX]: Add useFormErrors hook tests

### LINKED ISSUE #957: Add unit tests for the useFormErrors hook
## Problem
`apps/client/src/hooks/useFormErrors.ts` is the error-state hook behind every form in the app (login, users, dashboards, integrations…). It has no tests of its own — only its dependency `lib/errorMapper` does (#924), so the hook's branching is unverified.

## What to test
New file `apps/client/src/test/useFormErrors.test.ts` using `renderHook` from `@testing-library/react` (pattern: `apps/client/src/test/useCopyToClipboard.test.ts`):
- a successful response clears both `fieldErrors` and `generalError`
- a validation error response (`isValidationError`) populates per-field messages via `mapValidationErrors`
- with `showFieldErrors: false` (the login form) the same validation response yields one `generalError` and no field entries
- a non-validation failure routes through `mapApiError` into `generalError`
- `clearErrors()` resets everything

*(easy — build on the tested errorMapper primitives from #924; mock nothing, feed real response shapes)*

Rayan-and-beyond:
## What changed

- add direct hook coverage for successful responses clearing errors
- cover validation field mapping and the login general-error path
- cover general API error mapping and `clearErrors()`

## Validation

- `git diff --check`
- focused Vitest was attempted, but this disk-constrained sparse checkout cannot resolve the workspace `@OpsiMate/shared` package without reinstalling the broader workspace; the follow-up workspace install was stopped when free disk fell to ~622 MiB rather than pressure the shared VM

Closes #957

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for form error handling, including validation errors, authentication failures, successful submissions, and clearing previous errors.
  - Added checks for field-specific and general error states to help ensure consistent feedback during sign-in and form interactions.


idanlodzki:
Reviewed and verified locally on top of main: tests, lint, Prettier and the client typecheck gate all clean; CodeRabbit had no findings. All five behaviors from #957 pinned, with real response shapes rather than mocks — that's the coverage we wanted. Merging — thanks! If you like the project, a ⭐ on the repo helps us a lot.

PR #971, opened 2026-09-16
[FIX]: Restore forgot password rollback on mail failure

### LINKED ISSUE #767: forgotPassword rollback never runs because the mail send is fire-and-forget
## Problem
In `apps/server/src/bl/users/user.bl.ts` (lines ~123-142), the `try` block is intended to delete the just-created reset token if the email fails, but the send is `void this.mailClient.sendMail(...)` — not awaited — so a failure can never reach the `catch`, and the rollback (line ~140) is dead code.

Additionally `MailClient.sendMail` (`apps/server/src/dal/external-client/mail-client.ts` ~194-196) swallows its own errors, so it must rethrow for any caller to notice a failure.

## Suggested fix
- Await the `sendMail` call
- Let `sendMail` propagate its error (or return a success flag)
- Verify the token is deleted when the send fails

Swayamcodes:
## Issue Reference

Closes #767

---

## What Was Changed

* Await password-reset email delivery so failures reach the existing rollback path.
* Propagate errors from `MailClient.sendMail`.
* Handle rejected welcome-email promises while keeping registration non-blocking.
* Roll back only the token created by the failed request, preserving newer tokens from concurrent requests.
* Add regression coverage for mail failure propagation, token rollback, concurrent reset requests, and registration behavior.

---

## Why Was It Changed

The password-reset email was sent without awaiting its promise, and `MailClient.sendMail` swallowed delivery errors. As a result, the rollback never ran and failed email delivery could leave an unusable reset token stored.

---

## Screenshots

Not applicable—this is a server-side change.

---

Validation completed:

* Focused regression tests: 4 passed
* Full server suite: 378 passed
* Changed-file ESLint: passed
* Changed-file Prettier: passed
* Workspace build: 4 tasks passed

The repository-wide server lint currently reports an unrelated pre-existing `@typescript-eslint/no-unsafe-assignment` error in `apps/server/src/api/health.ts:22`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved password reset reliability when email delivery fails.
  - Failed password reset emails now show a clear error message and remove unusable reset requests.
  - Newer password reset requests remain valid when an earlier email attempt fails.
  - Registration can complete successfully even if the welcome email cannot be delivered.
  - Email delivery errors are now handled consistently instead of being silently ignored.


idanlodzki:
Reviewed and verified on top of current main: full server suite 378/378, lint at CI strictness, Prettier clean. The fix is right — awaiting the send and letting `MailClient.sendMail` propagate is exactly what #767 needed, and rolling back **only the token from this request** (not `deleteByUserId`) is a better answer than the issue asked for; the concurrent-request test that pins it is the best part of the PR. I also checked the one other `sendMail` caller that now sees a rejection (`playground.bl.ts` book-demo) — it already wraps the call in try/catch, so nothing else changes behavior.

Two small things to align with the repo's conventions before this can merge:
1. **Named interfaces, tests included** — this is enforced org-wide by CodeRabbit, so it only surfaces in review: the `{ token_hash: string }` cast and the `{ token?: string }` mock parameter in `forgotPassword.test.ts` should be declared interfaces (e.g. `PasswordResetTokenRow`, `SentMailOptions`).
2. **Import paths without `.ts`** — every other server test imports `'../src/…'` extensionless; please match (6 imports).

On the note in your description: I couldn't reproduce a pre-existing lint error in `health.ts:22` — `eslint src --ext .ts --max-warnings=0` is clean on main and on your branch, and CI's lint step passed. If your local run still shows it, it may be a stale `packages/shared/dist` (`pnpm --filter @OpsiMate/shared build`).

Push those two and it's ready; merge happens once the maintainer signs off. Thanks for a careful fix — and if you like the project, a ⭐ on the repo helps us a lot.

idanlodzki:
To save you a round-trip I applied the two convention items myself on your branch (803b0581 — named `PasswordResetTokenRow` / `SentMailOptions` interfaces, extensionless imports; test still 4/4). The maintainer has cleared this for merge; landing as soon as CI is green on that commit. Thanks again — and a ⭐ on the repo helps us a lot if you like the project.

Swayamcodes:
> To save you a round-trip I applied the two convention items myself on your branch (803b0581 — named `PasswordResetTokenRow` / `SentMailOptions` interfaces, extensionless imports; test still 4/4). The maintainer has cleared this for merge; landing as soon as CI is green on that commit. Thanks again — and a ⭐ on the repo helps us a lot if you like the project.

Thanks! I'll take care of that next time

PR #984, opened 2026-09-17
[FIX]: Add matcher column storage tests

### LINKED ISSUE #972: Add unit tests for parseMatcherColumn / serializeMatcherColumn
## Problem
`apps/server/src/dal/matcherColumn.ts` reads and writes the `label_matchers` column for mute policies and actions in two shapes — a legacy flat list and the newer OR-of-AND groups — and has no tests. `tests/nameMatchers.test.ts` covers only its siblings `parseNameColumn` / `serializeNameColumn`.

## What to test
New file `apps/server/tests/matcherColumn.test.ts`:
- `parseMatcherColumn`: `null`, `''` and `'[]'` all yield empty `labelMatchers` and `labelMatcherGroups`
- a flat list → `labelMatchers` is that list and `labelMatcherGroups` is `[list]`
- nested groups → `labelMatcherGroups` is the nesting and `labelMatchers` mirrors group 0
- malformed JSON → empty (never throws)
- `serializeMatcherColumn`: one group → flat JSON; several groups → nested; no groups → `[]`
- `parse(serialize(x))` round-trips for both shapes

*(easy — pure functions, no DB; mirror the style of `nameMatchers.test.ts`)*

ryux1:
Closes #972

## Summary
- cover empty, legacy flat, grouped, and malformed matcher-column input
- verify flat and grouped serialization shapes
- pin round trips for both supported storage formats

This is a tests-only change; production behavior is unchanged.

## Validation
- `pnpm --filter @OpsiMate/server exec vitest run tests/matcherColumn.test.ts` (11 passed)
- `pnpm --filter @OpsiMate/server test` (401 passed)
- `pnpm run test` (all workspace suites passed, including 497 client tests)
- `pnpm run check`
- `pnpm run build`

## Summary by CodeRabbit

* **Tests**
  * Added coverage for matcher-column parsing and serialization.
  * Verified handling of empty, malformed, legacy, and grouped matcher data.
  * Added round-trip checks to ensure matcher data preserves its structure.
  * Confirmed single-group data remains compatible with the legacy format.
  * Confirmed multiple groups are preserved when serialized and parsed.


idanlodzki:
Reviewed and verified on top of current main: 11/11 pass, Prettier clean, and it passes ESLint at the strictest setting we use (which is stricter than what CI runs on tests). CodeRabbit had nothing to say either.

This is exactly what #972 asked for — every bullet is pinned: the three empty inputs, the legacy flat list, nested groups (including the "mirrors group 0" contract for legacy consumers), malformed JSON never throwing, all three serialize shapes, and both round-trips. The shared `environmentMatcher`/`serviceMatcher` fixtures with an explicit `op` on one of them is a nice touch — it means the round-trip tests prove `op` survives storage, not just `key`/`value`. Opened about an hour after the issue went up, too.

We want to merge this — it'll land as soon as the maintainer signs off, which should be very soon. Thanks for a clean first contribution! If you like the project, a ⭐ on the repo genuinely helps us reach more people.

ryux1:
Thanks for verifying the full matrix and updating the branch against current main — glad the coverage matches what you had in mind for #972.

idanlodzki:
One small ask before this lands, @ryux1 — could you ⭐ star the repo? https://github.com/OpsiMate/OpsiMate

It takes a second and it's the single thing that most helps a small open-source project get found. Thanks — and welcome aboard, there are plenty more `good first issue`s if you want another.

idanlodzki:
Merging — thanks again, and welcome aboard. 🎉

PR #992, opened 2026-09-17
[FIX]: Validate numeric path parameters

### LINKED ISSUE #977: Non-numeric :id path params return 500 instead of 400 on five routers
## Problem
`GET /api/v1/tags/abc` answers **500 "Internal server error"** (verified live). Same for actions, enrichments, mute-policies and dashboards, and for custom-fields.

Cause: the five `*IdSchema`s in `packages/shared/src/schemas.ts` (around lines 44, 170, 291, 367, 441) do

```ts
z.string().transform((v) => { const n = parseInt(v, 10); if (isNaN(n)) throw new Error(...); return n; })
```

With zod 4, a `throw` inside `.transform()` surfaces as a plain `Error`, not a `ZodError`, so the controllers' `isZodError` branch is skipped and they fall through to 500. `parseInt('12abc')` is also silently `12`, so `PATCH /tags/12abc` targets tag 12. `apps/server/src/api/v1/custom-fields/controller.ts` (lines 66, 97, 138) has the same `parseInt` pattern inline.

## What to do
- add one helper in `schemas.ts`, e.g. `const idParam = (name: string) => z.object({ [name]: z.coerce.number().int().positive() })` — or keep the per-schema objects but use `z.coerce.number().int().positive()` instead of the transform
- use it in all five schemas and in the custom-fields controller
- one supertest case per router asserting `/abc` → 400 and `/12abc` → 400 (models: `tests/tags.test.ts`)

*(easy — the change is mechanical; the value is the six 400-tests that lock it in)*

ryux1:
## Issue Reference

Closes #977

---

## What Was Changed

Introduced one strict positive-integer ID schema for tags, dashboard tags, mute policies, enrichments, actions, and custom fields. Added integration coverage showing that both non-numeric and partially numeric values return 400 across all six routes.

---

## Why Was It Changed

`parseInt` accepts prefixes such as `12abc`, and inconsistent parsing allowed malformed path parameters to fall through to server errors.

---

## Screenshots

Not applicable; this is an API validation change.

---

Validated with the full test, lint, format, and build gates. Lint reports the repository's existing warnings and no errors.


idanlodzki:
Reviewed carefully since this touches shared schemas used by five routers: one `positiveIntegerIdSchema` on `z.coerce.number().int().positive()`, the custom-fields controller moved onto it with proper zod → 400 mapping, and a table-driven test proving `abc` and `12abc` are 400 on all six routes. Full server suite passes on top of main, lint at CI strictness, Prettier clean, CI green. One note for the record, not a blocker: `z.coerce` accepts `1e2` → 100 and `0x1f` → 31 — harmless (they still address a real integer id) and consistent with how Express query coercion behaves elsewhere. Merging — this closes a real bug (#977). Thank you!

PR #993, opened 2026-09-17
[FIX]: Make sidebar community links accessible

### LINKED ISSUE #983: Sidebar Slack and GitHub links are clickable divs — not focusable, not keyboard-activatable
## Problem
In `apps/client/src/components/LeftSidebar.tsx` (around lines 168–179 and 196–207) the Slack and GitHub buttons are

```tsx
<div className="… cursor-pointer …" onClick={() => window.open('https://…', '_blank')}>
```

A `div` with `onClick` has no role, is not in the tab order, and cannot be activated with Enter/Space — keyboard and screen-reader users cannot reach the community links at all.

## What to do
- replace each with a real link: `<a href="…" target="_blank" rel="noopener noreferrer" aria-label="Join our Slack">` (keep the icon inside; drop `window.open`)
- keep the existing classes so the visual stays identical
- extend `apps/client/src/test/iconButtonAccessibility.test.tsx` (or add a small test) asserting both render as `link` roles with accessible names

*(trivial — two elements; the value is that it becomes the pattern for future icon links)*

ryux1:
## Issue Reference

Closes #983

---

## What Was Changed

Replaced the clickable Slack and GitHub `div` elements with named external links and added accessibility coverage for their roles, names, targets, and safe `rel` values.

---

## Why Was It Changed

Native links are keyboard-focusable and keyboard-activatable without recreating link behavior in JavaScript.

---

## Screenshots

Not applicable; the visual presentation is unchanged.

---

Validated with the full test, lint, format, and build gates. Lint reports the repository's existing warnings and no errors.


idanlodzki:
Verified locally (5/5 incl. the new link assertions, lint, Prettier, typecheck) and CI is green. Real `<a>` elements with names, target and rel — this is now the pattern for icon links. Merging, thanks!

Issue #982, opened 2026-09-17
Export MOBILE_BREAKPOINT instead of hardcoding "768px" in the mobile overlay copy


idanlodzki:
## Problem
`apps/client/src/hooks/use-mobile.tsx:3` declares `const MOBILE_BREAKPOINT = 768` but does not export it, and `apps/client/src/components/MobileWebOverlay/MobileWebOverlay.tsx:35` hardcodes the same number in user-facing text: "Minimum recommended width: 768px". Changing one silently makes the other lie.

## What to do
- `export const MOBILE_BREAKPOINT = 768` from `use-mobile.tsx`
- interpolate it in the overlay copy: `` `Minimum recommended width: ${MOBILE_BREAKPOINT}px` ``
- if #975-style tests exist for `useIsMobile` by then, assert the query string uses the constant

*(trivial)*

idanlodzki:
Fixed by #995 (the PR description didn't carry a `Closes #982` line, so closing manually).

PR #995, opened 2026-09-18
fix: export MOBILE_BREAKPOINT and reuse it in MobileWebOverlay

Prash2185:
## Problem
`apps/client/src/hooks/use-mobile.tsx` declares `MOBILE_BREAKPOINT = 768`
but doesn't export it. Meanwhile,
`apps/client/src/components/MobileWebOverlay/MobileWebOverlay.tsx`
hardcodes the same value in its user-facing text:
"Minimum recommended width: 768px". Since the two values aren't linked,
changing one silently makes the other lie.

## Solution
- Exported `MOBILE_BREAKPOINT` from `use-mobile.tsx`
- Imported it in `MobileWebOverlay.tsx` and interpolated it into the
  overlay's copy instead of the hardcoded `768`
- Now both the breakpoint logic and the displayed text always stay
  in sync — changing the constant in one place updates both

## Summary by CodeRabbit

- **Bug Fixes**
  - The mobile web overlay now consistently displays the minimum recommended width of 768px, matching the app’s mobile breakpoint setting.
  - This ensures the guidance shown to users stays synchronized with the application’s mobile layout behavior across supported screen sizes.


Prash2185:
image

Screenshot confirming the fix — the overlay text now reads from
the exported MOBILE_BREAKPOINT constant instead of a hardcoded value:

idanlodzki:
Thanks — this is the right change (export the constant, interpolate it) and it came in first for #982. CI tripped only on formatting, so I pushed a Prettier-only commit on top of your branch rather than bounce it (`pnpm --filter @OpsiMate/client format-fix` before pushing next time and you'll never hit it). Merging as soon as CI is green on that commit. If you like the project, a ⭐ on the repo helps us a lot.

PR #998, opened 2026-09-19
[FIX]: Add unit tests for alert history range filtering and entry selection

### LINKED ISSUE #958: Add unit tests for alertHistory.utils range filtering and the synthesized-update fallback

chiliec:
## Issue Reference

Closes #958

---

## What Was Changed

- New `apps/client/src/test/alertHistorySelection.test.ts` covering the two untested exports of `alertHistory.utils.ts`:
  - `filterHistoryByRange`: entries exactly on `from`/`to` are kept (inclusive), entries 1 ms outside either bound are dropped, open-ended ranges (only `from` / only `to`) work, and `null`/`undefined`/empty ("All time") ranges return the input unchanged.
  - `selectHistoryEntries`: real events win when any fall inside the window; the synthesized last-`UPDATED` entry is returned as a fallback only when the window hides every real event; nothing is returned when the window hides the update too; on "All time" the fallback is never used even when there are no real events.
- `alertHistory.utils.ts`: the "All time" case from the issue did not hold when the history contained only the synthesized `UPDATED` entry (no real events at all) — `selectHistoryEntries` returned the `UPDATED` entry because it decided the fallback on `realEntries.length === data.length` rather than on whether a window is active. The window resolution is now a small shared helper (`resolveWindow`) used by both `filterHistoryByRange` and the fallback check, so "All time" always returns just the real events.

---

## Why Was It Changed

Issue #958 asks for unit tests over these pure functions. Writing the "on All time the fallback is never used even if there are no real events" case surfaced the mismatch above, so the test pins the documented behaviour and the function is adjusted to meet it. Behaviour for every other case (windows containing real events, windows hiding all real events) is unchanged.

---

## Screenshots

N/A — no UI change.

---

Verified locally (Node 22, pnpm 9.12.0):

- `pnpm exec vitest run src/test/alertHistorySelection.test.ts` on unmodified `main`: 8 passed, 1 failed (`never uses the fallback on "All time"` — received the `UPDATED` entry instead of `[]`).
- With the fix: `pnpm exec vitest run` in `apps/client` — 53 files / 516 tests passed.
- `pnpm typecheck` — no new client type errors vs. baseline; `eslint` and `prettier --check` clean on both touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved alert history filtering for bounded, open-ended, and unrestricted time ranges.
  - Ensured range boundaries are handled inclusively and entries outside the selected period are excluded.
  - Corrected update-event fallback behavior when relevant history is hidden by a selected time range, including All time views.
  - Preserved complete history when no time range is applied.

- **Tests**
  - Added coverage for range filtering, fallback scenarios, empty results, and All time behavior.


idanlodzki:
Thank you — this is careful work, and the `filterHistoryByRange` coverage (both bounds, ±1 ms, one-sided ranges, the three All-time inputs, same-reference return) is exactly what #958 asked for. I want to merge it, with one change.

**Please drop the source change in `alertHistory.utils.ts` and adjust the two All-time tests.** You implemented the contract as the code comment (and, honestly, my issue text) stated it — "never on All time" — but that comment is wrong, and checking it against the server showed why:

- The `UPDATED` entry is synthesized server-side from `updated_at` and is the **only** entry an alert has once its status-history rows are gone — which Data Retention does by default after 90 days. For a long-lived alert on "All time", today's code shows "Last update received…"; with `!resolveWindow(timeRange)` it would show an empty timeline. That's a user-visible regression hiding inside a tests PR, so I'd rather not take it.
- The condition that's actually there — `realEntries.length === data.length` ("if there are no synthesized entries, there's nothing to fall back to") — is the intended behavior; only the prose was off.

Concretely:
1. Revert `selectHistoryEntries` to main (keeping the small `resolveWindow` helper for `filterHistoryByRange` is fine if you like it; it's behavior-neutral there).
2. Change the two All-time assertions to pin the real contract: with real events present, All-time returns only the real events (the fallback stays out); with **only** an `UPDATED` entry, All-time returns that entry (`[updatedInside]`, not `[]`).
3. Fix the comment above `selectHistoryEntries` so the next reader isn't misled: the fallback is used whenever the current window shows no real events *and* a synthesized entry exists — including All-time.

I'll update #958's wording to match. Push that and I'll merge the same day. And if you like the project, a ⭐ on the repo helps us a lot.

chiliec:
Done in 93aa081 — you're right, the Data Retention case makes the "never on All time" reading wrong:

1. `selectHistoryEntries` is back to main's condition (`realEntries.length === data.length`); `resolveWindow` stays only as a helper for `filterHistoryByRange`, where it's behaviour-neutral.
2. The All-time tests now pin the real contract: with real events present only the real events are returned; with only an `UPDATED` entry, All-time returns `[updatedInside]`.
3. Comment above `selectHistoryEntries` rewritten: the fallback applies whenever the window shows no real event and a synthesized entry exists — including All time — and mentions the retention-pruned case.

`vitest run src/test/alertHistorySelection.test.ts`: 10 passed; `tsc --noEmit` and eslint clean on both files.

idanlodzki:
Exactly right — thank you for taking the correction so well and for writing the Data Retention rationale into the comment; the next reader will thank you. Verified on top of main: 10/10, lint, Prettier, typecheck gate; source diff is now comment + a behavior-neutral helper. CI green, CodeRabbit clean. Merging. If you like the project, a ⭐ on the repo helps us a lot.

PR #999, opened 2026-09-19
test: add AlertHistoryRepository tests

### LINKED ISSUE #979: Add unit tests for AlertHistoryRepository
## Problem
`apps/server/src/dal/alertHistoryRepository.ts` stores the per-alert timeline (ownership, silence, comment events…) and is only exercised indirectly through API tests. Its timestamp format matters: the history index (#897) relies on ISO-8601 UTC, and a regression to SQLite's `CURRENT_TIMESTAMP` would silently defeat it.

## What to test
New file `apps/server/tests/alertHistoryRepository.test.ts` (call `setupExpressApp(db)` once so the history table exists, then use the repository directly):
- `recordEvent` writes `created_at` as ISO-8601 UTC ending in `Z`
- missing `actorName` / `description` are stored as `null`
- `getEvents(alertId)` returns newest first and only that alert's rows
- `getEventTimesByType([ids], type)` groups timestamps per alert, filters by type, and returns `{}` for `[]`
- `getAllEventTimes` returns the (alert_id, created_at, actor_name) triples across alerts

*(easy — DAL only, no HTTP)*

AnuragR3k:
## Summary

- add unit tests for `AlertHistoryRepository`
- cover event recording, timestamps, nullable fields, ordering, filtering, grouping, and empty results
- link this PR to #979

## Validation

- `pnpm --filter @OpsiMate/server test -- alertHistoryRepository.test.ts` ✅ 8 tests passed
- `pnpm --filter @OpsiMate/server check` ✅

Closes #979

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive coverage for alert history behavior, including event recording, ISO-8601 timestamps, nullable values, filtering by alert, newest-first sorting, and empty-result handling.
  - Verified grouped event-time results preserve each event’s persisted timestamp.
  - Confirmed retrieved history includes actor names and that alert history data remains consistent when using SQLite storage.


idanlodzki:
Thanks — solid coverage of every bullet in #979 (ISO `Z` timestamps, null actor/description, newest-first, per-alert scoping, grouping, empty input, the all-events triples), and you used the setup helpers the right way. Verified locally: 8/8 on top of main, Prettier clean, lint clean at the strict setting.

One small thing before merge, from CodeRabbit's inline comment and worth doing: the grouping and all-events tests assert **membership and counts but not the timestamps themselves**, so a query that returned the wrong `created_at` values would still pass. Capture the `created_at` of each recorded event (via `getEvents`) and assert `getEventTimesByType` / `getAllEventTimes` return exactly those values. A few lines — push that and I'll merge the same day. And if you like the project, a ⭐ on the repo helps us a lot.

AnuragR3k:
Made the timestamp assertions change you suggested and pushed it in 5c67dc6.

Also ran the server tests, formatting and lint, everything is passing now. Thanks for the review! 🙌

idanlodzki:
Thanks for the quick turnaround — the timestamp assertions now prove `getEventTimesByType` and `getAllEventTimes` return the persisted values, not just the right counts. I pushed one tiny lint commit on top (comparing the stored `event_type` string against the enum value explicitly — `no-unsafe-enum-comparison` at the strict setting). 8/8 on top of main, Prettier, CI green. Merging; this closes #979. Thanks for the ⭐ as well — I noticed. 🙌

PR #1021, opened 2026-09-22
[FEAT]: Add enrichment version history

### LINKED ISSUE #748: Enrichment version history: save and view previous versions of enrichments
## Problem
Today we only keep an audit record of *when* a user made an enrichment. The actual content of previous versions is not saved, so there is no way to see what an enrichment looked like before it was changed.

## Proposed solution
Save a version of the enrichment content on every change, so users can view the full history of an enrichment — not just that a change happened, but what each version contained.

## Acceptance criteria
- [ ] Every enrichment change stores a new version (content, author, timestamp)
- [ ] Users can view the version history of an enrichment
- [ ] Users can see the content of each previous version

vansh-nagar:
Closes #748.

## What changed
- store immutable enrichment snapshots on create and update
- backfill existing enrichments as version 1 during initialization
- expose version history through the server API and shared types
- add a client history dialog with readable version details
- invalidate history data after edits and support the playground mock API

## Tests
- 4 focused server tests covering snapshots, no-op updates, missing records, and legacy backfill
- client, shared, custom-actions, and server builds
- client and server lint
- full server suite: 419 tests passed; the unrelated tags suite timed out once in the full run and passed 7/7 when rerun alone

## Summary by CodeRabbit

* **New Features**
  * Added version history tracking for alert enrichments.
  * Enrichment records now save snapshots when created or updated.
  * Added a history action on the Enrichments page with a scrollable dialog showing version details, authors, timestamps, and configuration changes.
  * Added API support for retrieving enrichment history, including clear not-found responses.
* **Bug Fixes**
  * Existing enrichments receive an initial history snapshot automatically.


idanlodzki:
Thanks for taking on #748 — this is a real feature, not a first issue, and the server side is in good shape: immutable snapshots on create and update inside a transaction, a backfill so existing rules have a "before" for their next edit, the history route registered ahead of `/:enrichmentId`, and tests for the four cases that matter. I read the whole diff.

Because it adds a table and a UI surface, the maintainer will look at it himself before it merges. Meanwhile, here's what would make that review a yes:

**CodeRabbit's five threads are all valid — please address them:**
1. `EnrichmentHistoryDialog` into its own directory with `index.ts`; `VersionContent` into its own file with a named `VersionContentProps` (the repo's named-interfaces rule applies everywhere).
2. Early returns for the loading / error / list states rather than a nested ternary.
3. `useEnrichmentHistory` in its own file, exported from the barrel.
4. Playground mock: keep snapshots on create/update so the dialog actually shows history after an edit in demo mode (right now it always returns one "version 1").
5. Backfill: persist the legacy row's `updated_at` as the version's `created_at`, not "now".

**Two things CodeRabbit didn't say:**
6. **Timestamp format.** `created_at DATETIME DEFAULT CURRENT_TIMESTAMP` writes SQLite's `YYYY-MM-DD HH:MM:SS`. Every table added this year writes ISO-8601 UTC from the app instead (see `alert_root_causes` / `rootCauseRepository.ts` for the pattern) — mixed formats are what forced the `datetime()` wrapper in retention and defeated its indexes. Please write `new Date().toISOString()` explicitly (and the legacy `updated_at` via `toIsoUtc` for the backfill).
7. **Growth.** Versions are unbounded — every save adds a full JSON snapshot. Either add `EnrichmentVersions` to `RetentionResource` with a policy row (the retention repo already handles per-resource cutoffs), or at minimum note in the PR that it's deferred, so the maintainer can decide.

Also: a screenshot of the dialog from `?playground` in the description would help the review a lot. And if you like the project, please ⭐ star the repo — https://github.com/OpsiMate/OpsiMate — it genuinely helps.

PR #1022, opened 2026-09-22
fix: trim alert search terms before matching

### LINKED ISSUE #1018: Alert search with trailing whitespace matches nothing
## Problem
`searchAlerts` in `packages/shared/src/alerts/alertQuery.ts` (around line 113) trims the term for the empty check but matches with the **untrimmed** value:

```ts
if (!searchTerm.trim()) return alerts;
const lower = searchTerm.toLowerCase();   // not trimmed
```

So typing `high ` (with a trailing space — easy to do when pasting) matches nothing against "cpu high". Every other search box in the client trims (Actions, On-call, Mute Policies).

## What to do
- `const lower = searchTerm.trim().toLowerCase();`
- add a case to `apps/server/tests/alertQueryEngine.test.ts` (where the shared engine is tested): `"high "` and `" high"` both match an alert named "cpu high"

*(trivial, user-visible)*

theman6660:
Fixes #1018

## Summary

`searchAlerts` returned every alert for a blank query by checking `searchTerm.trim()`, then matched with `searchTerm.toLowerCase()` without trimming. `high ` and ` high` therefore missed an alert named `cpu high`. The other search boxes in the client already trim.

The shared engine now trims once and matches on that value. A whitespace-only query still returns the original list.

## Verification

Imported the updated `searchAlerts` and checked:

- `high ` and ` high` both return the alert named `cpu high`
- a whitespace-only query returns the same array it was given

`apps/server/tests/alertQueryEngine.test.ts` covers the same cases. I did not run the full server Vitest suite, because that pulls in the Express app and SQLite setup.

## Summary by CodeRabbit

- **Bug Fixes**
  - Alert searches now ignore leading and trailing whitespace.
  - Searches containing only whitespace return all available alerts.
  - Case-insensitive matching continues across alert names, statuses, tags, summaries, comments, and integrations.


idanlodzki:
Verified on top of main (shared rebuilt, 12/12 in `alertQueryEngine.test.ts`, lint at CI strictness, Prettier) and CI is green. Trim once, use it for both the empty check and the match — and the whitespace-only case pinning the same-reference return is a nice extra. Merging; this closes #1018. Two asks: if you like the project, please ⭐ star the repo — https://github.com/OpsiMate/OpsiMate — it genuinely helps a small project get found; and there are more `good first issue`s open if you want another.

PR #1023, opened 2026-09-22
[FIX]: Cover dashboard tag endpoints

### LINKED ISSUE #1014: Add unit tests for the dashboard tag endpoints
## Problem
`apps/server/src/api/v1/dashboards/controller.ts` has four tag handlers — attach, detach, list per dashboard, list across dashboards — and `tests/dashboards.test.ts` covers only column settings (the only tag reference in the suite is #992's 400-on-bad-id check).

## What to test
New file `apps/server/tests/dashboardTags.test.ts` (`setupDB`, `setupExpressApp`, `setupUserWithToken`; create a dashboard and a tag through the API first):
- `POST /api/v1/dashboards/:id/tags` with an unknown dashboard → 404 "Dashboard not found"; with an unknown `tagId` → 404 "Tag with ID n not found"
- happy path, then `GET /api/v1/dashboards/:id/tags` lists the tag
- `DELETE /api/v1/dashboards/:id/tags/:tagId` removes it; a second delete is still 200 (idempotent)
- `GET /api/v1/dashboards/tags` aggregates tags per dashboard

*(easy — model on `tests/tags.test.ts`)*

vansh-nagar:
Adds API-level regression coverage for all four dashboard-tag routes:

- missing dashboard and tag responses
- attach and per-dashboard listing
- idempotent detach behavior
- cross-dashboard tag aggregation

The tests create their fixtures through the public API and assert the response contract end to end.

Checks run: focused Vitest suite (4 tests) and Prettier.

Fixes #1014

## Summary by CodeRabbit

* **Tests**
  * Added coverage for Dashboard Tags API behavior, including:
    * Handling missing dashboards and tags.
    * Adding, listing, and removing dashboard tags.
    * Repeated tag deletion.
    * Retrieving dashboards with their associated tags.


idanlodzki:
Thanks — this is close. Verified on top of main: 4/4, Prettier clean, and the fixture approach (fresh dashboard + tag per test) is the right way to keep the cases independent. Two small things before merge:

1. **CodeRabbit's inline comment is right**: the aggregation test attaches a tag to one dashboard, so a route that dropped or mis-attributed other dashboards' tags would still pass. Create a second dashboard + tag in that test and assert both groups come back correctly.
2. The repo has a **named-interfaces rule** (no anonymous inline object types, tests included — it's enforced org-wide by CodeRabbit so it only shows up in review): `(item: { dashboardId: number })` in the aggregation test should be a small declared interface, e.g. `DashboardTagGroup`.

Push those and I'll merge the same day. And if you like the project, please ⭐ star the repo — https://github.com/OpsiMate/OpsiMate — it genuinely helps.

PR #1026, opened 2026-09-22
fix: trim alert search term before matching (#1018)

### LINKED ISSUE #1018: Alert search with trailing whitespace matches nothing

Evaloric-ZA:
Fixes #1018.

`searchAlerts` trimmed the term for the empty check but matched with the untrimmed value, so `"high "` (trailing space, easy when pasting) found nothing. Trim once and reuse, exactly as the issue suggests.

Tests: two cases added to `apps/server/tests/alertQueryEngine.test.ts` per the issue (`"high "` and `" high"` both match an alert named "cpu high`). Verified locally with `vitest run tests/alertQueryEngine.test.ts`: 13/13 pass with the fix; reverting only the source line fails the trailing-space case, confirming the guard bites. Full suite not run (single-file scope).

## Summary by CodeRabbit

- **Bug Fixes**
  - Alert searches now ignore leading and trailing spaces, producing expected matches when search terms contain accidental whitespace.

- **Tests**
  - Added regression coverage for searches with leading and trailing whitespace.


idanlodzki:
Thank you — this is correct and the two tests read well. Unfortunately #1022 fixed the same issue (#1018) identically and was opened two hours earlier, so I've merged that one and I'm closing this as a duplicate; first-come on identical fixes, nothing about your work. Thanks for the ⭐ too — I noticed. Plenty of `good first issue`s are open; comment on one to claim it and I'll review quickly.

